PR #1, opened 2026-08-26
feat: first-class API_KEY — serve keyed and authenticate the script's own readiness/status/smoke curls

sethforprivacy:
Deployed this recipe today on a 2x DGX Spark pair (works beautifully — thank you) with an API key via `EXTRA_ARGS=--api-key …`, and hit a footgun worth fixing at the source:

**A keyed server makes `start.sh` report failure on a successful boot.** Every readiness/status/smoke curl in the script uses `-f` with no Authorization header, so the server's 401s read as "not ready". `wait_ready` burned its full timeout and exited nonzero while the server had been up and serving for most of an hour. The containers survive (nothing tears down), but the operator signal is exactly wrong.

This PR makes `API_KEY=` a documented first-class env:

- appended to `SGLANG_ARGS` as `--api-key` (before `EXTRA_ARGS`, so explicit overrides still win)
- sent as a `Bearer` header by `wait_ready`, `status`, and `smoke` via an `AUTH_CURL` array
- empty `API_KEY` preserves the exact current behavior, byte for byte

Field-tested on our pair: keyed boot now reaches `ok "SGLang is ready"` and `./start.sh smoke` passes against the keyed server.

Second, unrelated QoL PR incoming for boot persistence.

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -138,6 +138,8 @@
 #    WAIT_TIMEOUT_MIN=90
 #    NCCL_DEBUG=WARN
 #    EXTRA_ARGS="…"             appended last (argparse last-wins → overrides)
+#    API_KEY=                   optional: serve with --api-key AND authenticate
+#                               this script's own readiness/status/smoke curls
 # =============================================================================
 set -euo pipefail
 
@@ -232,6 +234,12 @@ WAIT_TIMEOUT_MIN="${WAIT_TIMEOUT_MIN:-90}"
 NCCL_DEBUG="${NCCL_DEBUG:-WARN}"
 NCCL_CROSS_NIC="${NCCL_CROSS_NIC:-0}"
 EXTRA_ARGS="${EXTRA_ARGS:-}"
+API_KEY="${API_KEY:-}"
+# Auth header for every local readiness/status/smoke curl in this script:
+# with --api-key set, the server answers 401 to bare requests, which curl -f
+# treats as failure -- without this, wait_ready times out on a healthy server.
+AUTH_CURL=()
+[[ -n "${API_KEY}" ]] && AUTH_CURL=(-H "Authorization: Bearer ${API_KEY}")
 
 # ---- Paths / logs -----------------------------------------------------------
 HF_HOME="${HF_HOME:-${HOME}/.cache/huggingface}"
@@ -891,6 +899,7 @@ build_sglang_args() { # $1 = container model path
   else
     SGLANG_ARGS+=(--cuda-graph-backend-decode=disabled)
   fi
+  [[ -n "${API_KEY}" ]] && SGLANG_ARGS+=(--api-key "${API_KEY}")
   # Free-form overrides go last (argparse last-wins).
   read -ra _extra <<< "${EXTRA_ARGS}"
   if [[ ${#_extra[@]} -gt 0 ]]; then SGLANG_ARGS+=("${_extra[@]}"); fi
@@ -1047,7 +1056,7 @@ wait_ready() {
   local deadline=$(( $(date +%s) + WAIT_TIMEOUT_MIN * 60 )) start elapsed heartbeat
   start=$(date +%s); heartbeat=0
   while :; do
-    if curl -fsS "${READY_URL}" >/dev/null 2>&1; then
+    if curl -fsS "${AUTH_CURL[@]}" "${READY_URL}" >/dev/null 2>&1; then
       echo ""
       ok "SGLang is ready at ${READY_URL}"
       return 0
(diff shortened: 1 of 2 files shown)


mrexodia:
Isn't this more something you would handle in your proxy?

sethforprivacy:
> Isn't this more something you would handle in your proxy?

It depends, can be handled directly via SGLang, but can also be done via proxy. Main benefit to auth in vLLM/SGLang is that it is native bearer auth, while most proxies normally just expose basic auth or similar which won't work with many harnesses.

start.sh
```diff
@@ -891,6 +899,7 @@ build_sglang_args() { # $1 = container model path
   else
     SGLANG_ARGS+=(--cuda-graph-backend-decode=disabled)
   fi
+  [[ -n "${API_KEY}" ]] && SGLANG_ARGS+=(--api-key "${API_KEY}")
```
MiaAI-Lab, line 902: **[suggestion]** `--api-key` is added from `API_KEY` here, then `EXTRA_ARGS` is appended last (argparse last-wins). `AUTH_CURL` is built only from `API_KEY` (start.sh:242) and never inspects `EXTRA_ARGS`. Two realistic configs still reproduce the original footgun: (1) `EXTRA_ARGS=--api-key <key>` with `API_KEY` empty — serving is keyed, but `wait_ready`/`status`/`smoke` keep sending unauthenticated `-f` curls and treat 401 as “not ready”; (2) both set to different values — curls authenticate as `API_KEY` against a server that expects the `EXTRA_ARGS` key, so a healthy boot still times out. The PR body calls out `EXTRA_ARGS=--api-key` as the path that burned a 90-minute wait today.

**Suggestion:** If `EXTRA_ARGS` contains `--api-key` and `API_KEY` is empty, warn (and ideally parse that value into `AUTH_CURL`). If both are set, warn that `EXTRA_ARGS` wins for the server but curls still use `API_KEY`. Simplest complete fix: derive the effective key after `build_sglang_args` and build `AUTH_CURL` from that.

start.sh
```diff
@@ -232,6 +234,12 @@ WAIT_TIMEOUT_MIN="${WAIT_TIMEOUT_MIN:-90}"
 NCCL_DEBUG="${NCCL_DEBUG:-WARN}"
 NCCL_CROSS_NIC="${NCCL_CROSS_NIC:-0}"
 EXTRA_ARGS="${EXTRA_ARGS:-}"
+API_KEY="${API_KEY:-}"
+# Auth header for every local readiness/status/smoke curl in this script:
+# with --api-key set, the server answers 401 to bare requests, which curl -f
+# treats as failure -- without this, wait_ready times out on a healthy server.
+AUTH_CURL=()
+[[ -n "${API_KEY}" ]] && AUTH_CURL=(-H "Authorization: Bearer ${API_KEY}")
```
MiaAI-Lab, line 242: **[suggestion]** `AUTH_CURL` covers the script’s own curls, but the operator-facing “Quick test” printed after a successful `cmd_serve` (start.sh:1130–1131) still emits an unauthenticated `curl`. After a keyed boot that is the first command operators copy, and it 401s. README.md (unchanged in this PR) has the same unauthenticated example and no `API_KEY` row in its tunables table.

**Suggestion:** When `API_KEY` is set, print `-H "Authorization: Bearer ${API_KEY}"` on that example (or a placeholder that does not echo the secret, plus a note that `API_KEY` is required). Add `API_KEY` to the README tunables table and the documented curl.

start.sh
```diff
@@ -138,6 +138,8 @@
 #    WAIT_TIMEOUT_MIN=90
 #    NCCL_DEBUG=WARN
 #    EXTRA_ARGS="…"             appended last (argparse last-wins → overrides)
+#    API_KEY=                   optional: serve with --api-key AND authenticate
```
MiaAI-Lab, line 141: **[nit]** `API_KEY` is documented at the end of the header comment block, which `.env.example` points operators to via `./start.sh --help`. `--help` is `sed -n '2,120p'`, so these new lines (and already `EXTRA_ARGS` / `WAIT_TIMEOUT_MIN`) never appear in the help output.

**Suggestion:** Extend the `sed` range to cover the full header (through the closing `====` line), or generate help from a delimiter rather than a hardcoded line number, so newly documented tunables show up.

MiaAI-Lab (approved):
Approved. This is the right fix for the keyed-server footgun: `API_KEY` both passes `--api-key` and authenticates `wait_ready` / `status` / `smoke`. Empty `API_KEY` is unchanged.

Inline notes below are non-blocking follow-ups, not merge blockers:
- `EXTRA_ARGS=--api-key …` still keys the server without authenticating those curls — migrate the key to `API_KEY=`
- the post-serve “Quick test” curl (and README) still omit the Bearer header
- `--help` is `sed -n '2,120p'`, so the new header lines never print

MiaAI-Lab:
Merging this. Set `API_KEY=` in `.env` (or the environment) rather than `EXTRA_ARGS=--api-key …` — the latter still keys SGLang without authenticating `start.sh`'s own curls.

Thanks @sethforprivacy — this was a real production footgun.

PR #4, opened 2026-08-26
Use docker save and docker load with pipe

hnakamur:
Since this is much faster thanks to a direct ConnectX-7 200 Gb RoCEv2 link.

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -392,19 +392,11 @@ ensure_base_image() {
   if worker_docker image inspect "${BASE_IMAGE}" >/dev/null 2>&1; then
     ok "base image present on worker"
   else
-    info "base image missing on worker — trying 'docker pull' there…"
-    if worker_docker pull "${BASE_IMAGE}" >/dev/null 2>&1; then
-      ok "base image pulled on worker"
-    else
-      warn "worker pull failed — syncing image head → worker via docker save/load"
-      local tar="/tmp/qwen38-flash-next-image-$$.tar"
-      docker save "${BASE_IMAGE}" -o "${tar}"
-      scp "${SSH_OPTS[@]}" "${tar}" "${WORKER_SSH}:/tmp/qwen38-flash-next-image.tar"
-      rm -f "${tar}"
-      worker_docker load -i /tmp/qwen38-flash-next-image.tar
-      wrun "rm -f /tmp/qwen38-flash-next-image.tar"
-      worker_docker image inspect "${BASE_IMAGE}" >/dev/null 2>&1 || { error "image sync to worker failed"; exit 1; }
+    info "base image missing on worker — trying 'docker load' there…"
+    if docker save "${BASE_IMAGE}" | worker_docker load >/dev/null 2>&1; then
       ok "base image loaded on worker"
+    else
+      warn "worker load failed — syncing image head → worker via docker save/load"
     fi
   fi
 }
--- a/start.sh
+++ b/start.sh
@@ -392,12 +392,10 @@ ensure_base_image() {
   if worker_docker image inspect "${BASE_IMAGE}" >/dev/null 2>&1; then
     ok "base image present on worker"
   else
-    info "base image missing on worker — trying 'docker load' there…"
-    if docker save "${BASE_IMAGE}" | worker_docker load >/dev/null 2>&1; then
-      ok "base image loaded on worker"
-    else
-      warn "worker load failed — syncing image head → worker via docker save/load"
-    fi
+    info "base image missing on worker — streaming via docker save | docker load…"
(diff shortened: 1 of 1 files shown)


start.sh
```diff
@@ -392,19 +392,11 @@ ensure_base_image() {
...
-      scp "${SSH_OPTS[@]}" "${tar}" "${WORKER_SSH}:/tmp/qwen38-flash-next-image.tar"
-      rm -f "${tar}"
-      worker_docker load -i /tmp/qwen38-flash-next-image.tar
-      wrun "rm -f /tmp/qwen38-flash-next-image.tar"
-      worker_docker image inspect "${BASE_IMAGE}" >/dev/null 2>&1 || { error "image sync to worker failed"; exit 1; }
+    info "base image missing on worker — trying 'docker load' there…"
+    if docker save "${BASE_IMAGE}" | worker_docker load >/dev/null 2>&1; then
       ok "base image loaded on worker"
+    else
+      warn "worker load failed — syncing image head → worker via docker save/load"
```
MiaAI-Lab: **[bug]** If `docker save | worker_docker load` fails, the function only prints a stale warning ("syncing image head → worker via docker save/load") and returns. The old tar/scp/`docker load`/`image inspect` fallback (and its `exit 1` on inspect failure) was removed and not replaced. Because the failure is inside `if …; then`, `set -e` does not abort. With `KERNEL_PATCH=0` the worker later runs `BASE_IMAGE` and will fail; with the default `KERNEL_PATCH=1`, worker `docker build` may pull `FROM ${BASE_IMAGE}` as a silent, slower fallback, or fail if the registry is unreachable.

**Suggestion:** Treat pipe failure as fatal (restore the inspect-or-exit check), or implement a real second path. For example: `docker save "${BASE_IMAGE}" | worker_docker load` then `worker_docker image inspect "${BASE_IMAGE}" >/dev/null 2>&1 || { error "image sync to worker failed"; exit 1; }`. Drop the leftover warn that claims a sync is about to happen.

start.sh
```diff
@@ -392,19 +392,11 @@ ensure_base_image() {
...
-      warn "worker pull failed — syncing image head → worker via docker save/load"
-      local tar="/tmp/qwen38-flash-next-image-$$.tar"
-      docker save "${BASE_IMAGE}" -o "${tar}"
-      scp "${SSH_OPTS[@]}" "${tar}" "${WORKER_SSH}:/tmp/qwen38-flash-next-image.tar"
-      rm -f "${tar}"
-      worker_docker load -i /tmp/qwen38-flash-next-image.tar
-      wrun "rm -f /tmp/qwen38-flash-next-image.tar"
-      worker_docker image inspect "${BASE_IMAGE}" >/dev/null 2>&1 || { error "image sync to worker failed"; exit 1; }
+    info "base image missing on worker — trying 'docker load' there…"
+    if docker save "${BASE_IMAGE}" | worker_docker load >/dev/null 2>&1; then
```
MiaAI-Lab: **[suggestion]** `>/dev/null 2>&1` hides both `docker save` and remote `docker load`/SSH errors. Combined with the missing fatal check, a ~30 GB stream failure becomes a single misleading warn with no root cause (broken SSH stdin, SIGPIPE, daemon error, disk full on the worker).

**Suggestion:** Keep stdout quiet if desired, but leave stderr visible, e.g. `if docker save "${BASE_IMAGE}" | worker_docker load >/dev/null; then`.

MiaAI-Lab:
Pushed cdb9e79 onto this branch to address the review:

- Dropped the leftover `else` that warned about a tar/scp sync it no longer does
- `docker save | worker_docker load` is now fatal on failure (`set -e` + `pipefail`)
- `worker_docker image inspect` or `exit 1` after the stream
- Load stdout stays quiet; stderr from save/load/SSH is visible

Ready to merge from this side.

MiaAI-Lab (commented):
Addressed on the branch in cdb9e79: stream failure is now fatal, inspect-or-exit is restored, and save/load/SSH stderr is left visible.

start.sh
MiaAI-Lab: Fixed in cdb9e79. The leftover `else` is gone: a failed `docker save | worker_docker load` is now fatal (`set -e` + `pipefail`), then `image inspect` or `exit 1`. Stderr from save/load/SSH is left visible.

start.sh
MiaAI-Lab: Fixed in cdb9e79. The pipeline now redirects only stdout (`>/dev/null`), so `docker save` / remote `docker load` / SSH errors stay on stderr.

hnakamur: Thank you for your fix!
I'm closing this now.

PR #7, opened 2026-08-27
fix: PR #1 review follow-ups — effective API key for the script's curls, keyed examples, untruncated --help

sethforprivacy:
Implements all three of your non-blocking review suggestions from #1, plus what an adversarial review pass surfaced while validating them.

## 1. Effective-key derivation
`derive_effective_api_key()` scans the assembled `SGLANG_ARGS` (or `EXTRA_ARGS` pre-dispatch, covering `status`/`smoke` which never build args) for the **last** `--api-key`/`--api-key=` token — matching argparse last-wins — and rebuilds `AUTH_CURL` from it. Warns on both configs you called out: key via `EXTRA_ARGS` with `API_KEY` empty, and both set (EXTRA_ARGS wins server-side).

Validation: fuzzed **1,458 knob configurations** against an independent model of argparse's parse — **zero** cases where the server ends up keyed and the script derives a different key. (The only divergence class is `--api-key=` with an empty value, which un-keys the server while the script still sends a header — sglang ignores it; benign direction only.) Also confirmed against sglang 0.5.10 source that its own warmup self-auth (`http_server.py` `_wait_and_warmup`) uses the same last-wins value, so the derivation can't disagree with an internal caller — and that `/health*`+`/metrics*` are auth-exempt, so keyed servers keep working with Prometheus.

## 2. Keyed operator examples
The post-serve Quick test now prints the `Authorization` header when a key is in effect — as a **literal placeholder**, never the secret (byte-verified with a canary key): `$API_KEY` when that env supplies the key, `<your-api-key>` when it came via `EXTRA_ARGS` (so pasting it can't silently expand to empty). README: `API_KEY` row in the tunables table, header on the documented curl, plus a note on the agent-harness JSON block, which carries the same 401 footgun.

## 3. `--help` truncation
Boundary-based header extraction (POSIX sed, verified on GNU and BSD): the old `2,120p` range was silently dropping **20** documented tunables, not 3. The boundary also tolerates blank lines inside the header, so future edits can't re-introduce silent truncation.

Unkeyed behavior is provably unchanged: differential run of every action (`serve status stop smoke logs doctor download`) against the parent commit with stubbed docker/ssh/curl — identical output and exit codes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,11 +128,16 @@ Or just `./start.sh` — it runs doctor → download → serve in order.
 ```bash
 curl http://127.0.0.1:8888/v1/chat/completions \
   -H 'Content-Type: application/json' \
+  -H "Authorization: Bearer $API_KEY" \
   -d '{
         "model": "Qwen3.8-Flash-Next-NVFP4",
         "messages": [{"role": "user", "content": "Explain RoCEv2 in one paragraph."}]
       }'
 ```
+
+The `Authorization` header is only needed when you set `API_KEY` (see [Key
+tunables](#key-tunables)); on an open server you can drop that line. Keyed
+servers answer bare requests with `401`.
 
 Thinking is on by default (`reasoning_content` streams separately). To turn it
 off per-request:
@@ -174,6 +179,9 @@ Wiring it into a local agent harness (e.g. pi's `~/.pi/agent/models.json`):
   }]
 }
 ```
+
+That block assumes an open server; with `API_KEY` set, put the key in
+`"apiKey"` and flip `"authHeader": true` / `"auth": "bearer"`.
 
 ## Commands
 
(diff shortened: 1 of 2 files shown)


MiaAI-Lab:
Safe, and landed on `main` as `89414aa`.

GitHub could not auto-merge this PR (`mergeable_state: dirty` against current `main` — 1M YaRN, NVFP4 KV, issue #8 guards, `SGLANG_HOST_IP`). The three follow-ups are applied on top of that tree:

1. `derive_effective_api_key()` — last `--api-key` on the assembled line (including `EXTRA_ARGS`) drives `AUTH_CURL`, so keyed `wait_ready` / `status` / `smoke` no longer hang on 401.
2. Keyed Quick-test / README examples print an Authorization placeholder, never the secret.
3. `--help` prints the whole header comment block instead of `2,120p`.

Unkeyed path unchanged. Closing this PR as landed rather than merging the stale branch.

MiaAI-Lab:
Landed on main as 89414aa (conflicts with current main; equivalent applied there).

Issue #8, opened 2026-08-27
MEM_FRACTION_STATIC=0.70 double-counts the PLE table: KV pool at 0.52x the advertised context, max_running_requests silently 16 -> 5


sethforprivacy:
Running this recipe on a 2× DGX Spark (GB10) pair — 200 Gb/s direct DAC, MTU 9000, `.env` overriding only the fabric IPs/IFs, `HF_REVISION`, `IMAGE`, and `EXTRA_ARGS=--api-key …`. Every serve knob is therefore the script default.

The pair serves correctly, so this is a **sizing** report rather than a boot failure. Summary: the `MEM_FRACTION_STATIC=0.70` default appears to reserve the PLE table twice, which starves the KV and mamba pools by roughly 26 GB/rank. Raising it to 0.80 — the value `start.sh:41` already documents — gives **5.2× the KV pool** and lifts `max_running_requests` from an effective **5 to 16**, measured on the same hardware.

## Measured, before and after

Both boots are the same pair, same image, same checkpoint; only `MEM_FRACTION_STATIC` differs.

|  | `0.70` (default) | `0.80` |
|---|---:|---:|
| weight load, main model | 73.13 GB | **73.47 GB** |
| mamba cache | 28 slots (1.43 GB) | **143 slots** (5.78 GB) |
| `max_running_requests is capped …` warning | present → **5** | **absent** |
| effective `max_running_requests` | **5** | **16** (the requested value) |
| KV pool | 137,408 tok (1.58 GB) | **715,136 tok** (8.18 GB) |
| KV pool ÷ one `--context-length` request | **0.52×** | **2.73×** |
| decode CUDA graphs captured | `bs=[1,2,3,4,5]` | `bs=[1,2,3,4,5,6,7,8,10,12,14,16]` |
| `available_gpu_mem` after boot | **33.71 GB** | 20.62 GB |

The default boot, verbatim:

```
max_running_requests is capped to 5 by the mamba state cache
  (max_mamba_cache_size=28, 5 state slots per request).
KV Cache is allocated. dtype: torch.bfloat16, #tokens: 137408, K size: 0.79 GB, V size: 0.79 GB
max_total_num_tokens=137408, chunked_prefill_size=4096, max_prefill_tokens=16384,
  max_running_requests=5, context_len=262144, available_gpu_mem=33.71 GB
```

and at 0.80:

```
[...]

PR #10, opened 2026-08-27
fix: two keyed-server blind spots — kv-eval 401s, and the issue #8 pool guard silently skips

sethforprivacy:
Two keyed-server blind spots, both the same class as #7 and both in code added after it. Found while upgrading a 2× DGX Spark pair from `dccb035` to current `main`.

## 1. `kv-eval` cannot authenticate at all

`evals/nvfp4_kv_eval.py` has no `--api-key` and never sets an `Authorization` header, so on a keyed server every probe 401s and the run aborts immediately:

```
NVFP4 KV eval  2026-08-27T09:55:19Z
  endpoint  http://127.0.0.1:8888
API not reachable at http://127.0.0.1:8888: HTTP Error 401: Unauthorized
```

Fix: `--api-key` defaulting to `$API_KEY`, injected into `_http_json` — the single request path, so one hook covers every probe. `start.sh kv-eval` forwards `EFFECTIVE_API_KEY` (already derived at startup, so it also picks up a key that only exists in `EXTRA_ARGS`).

## 2. The issue #8 pool guard silently skips on keyed servers

`boot_health_check()` reads `/get_server_info` without `AUTH_CURL`. That endpoint requires auth; `/metrics` does not. So `pool` reads fine, `ctx` comes back `0`, the `(( pool > 0 && ctx > 0 ))` test fails, and the guard degrades to:

```
━━━ KV pool vs advertised context ━━━
[WARN]  could not read pool/context from the live API — skip size check
```

Measured on the same pair:

```
/get_server_info unauth -> 401
/get_server_info authed -> 200
/metrics         unauth -> 200
```

So the fail-closed check added for #8 never fires on a keyed deployment — and a keyed deployment is exactly what we were running when we hit #8 (a 137,408-token pool advertising 262,144, silently truncating a real client's 137.5K conversation). It fails *open* precisely where it matters.

Fix: pass `"${AUTH_CURL[@]}"`, matching how `wait_ready` and `cmd_status` already do it.

## Verification

On 2× DGX Spark (GB10 / SM121) TP2, 1M YaRN, `NVFP4_KV_CACHE=1`, five-key server:

- `./start.sh kv-eval` now runs: pool **1,826,752 tokens / 7.19 GB**, `nvfp4=True`, cases **11/11**, niah **7/7**, verdict **RELIABLE** — independently reproducing the 11/11 in your CHANGELOG.
- The pool guard now reports **1.74× one context-length request** instead of skipping.
- `bash -n start.sh` clean; `nvfp4_kv_eval.py --help` renders the new flag, and argparse shows the literal `$API_KEY` rather than a resolved secret.

`--api-key` on the eval takes the key on the command line, which is visible in `ps` for the duration of the run. `$API_KEY` from the environment is the quieter path and is the default; happy to make the flag env-only if you'd rather not offer the argument at all.

Unrelated but while I'm here: thanks for landing the #8 fix and #7 so fast — the 0.80 default took our pair from a 5-request admission cap to 28, and NVFP4 KV took the pool from 624K bf16 tokens to 1.83M for less memory.

### DIFF
--- a/evals/nvfp4_kv_eval.py
+++ b/evals/nvfp4_kv_eval.py
@@ -21,6 +21,7 @@ from __future__ import annotations
 import argparse
 import hashlib
 import json
+import os
 import random
 import re
 import sys
@@ -96,6 +97,11 @@ def _color(enabled: bool, code: str, text: str) -> str:
     return f"\033[{code}m{text}\033[0m"
 
 
+# Populated from --api-key / $API_KEY before the first request. A keyed server
+# 401s every probe otherwise, which surfaces as "API not reachable".
+_AUTH_HEADER: dict[str, str] = {}
+
+
 def _http_json(
     url: str,
     *,
@@ -106,7 +112,8 @@ def _http_json(
     req = urllib.request.Request(
         url,
         data=data,
-        headers={"Content-Type": "application/json", "Accept": "application/json"},
+        headers={"Content-Type": "application/json", "Accept": "application/json",
+                 **_AUTH_HEADER},
         method="GET" if payload is None else "POST",
     )
     return urllib.request.urlopen(req, timeout=timeout)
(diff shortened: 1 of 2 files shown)


MiaAI-Lab (approved):
Same class as #7, in the two places that missed `AUTH_CURL` after it.

- `boot_health_check` on `/get_server_info` now matches `wait_ready` / `status`. Unkeyed path is a no-op (`AUTH_CURL` empty). Keyed path actually runs the issue #8 pool guard instead of skipping.
- `--api-key` (default `$API_KEY`) into `_http_json` is the one hook; `./start.sh kv-eval` forwards `EFFECTIVE_API_KEY`. Fine leaving the flag — env is the quieter path, as you said.

Approving and merging.

PR #14, opened 2026-08-27
fix: verify worker weights by bytes, not just file count (a truncated shard passes today)

sethforprivacy:
## The defect

`sync_weights_to_worker()` accepts the worker's copy on a **file+symlink count**
match under `snapshots/`, plus `config.json` existing. That is the whole
acceptance test, on both the "already synced" early return and after the rsync:

```bash
hcount="$(find "$(snapshot_root)/snapshots" \( -type f -o -type l \) … | wc -l)"
wcount="$(wrun "find '${wrepo}/snapshots' \( -type f -o -type l \) … | wc -l")"
if [[ "${hcount}" == "${wcount}" ]] && wrun "test -f '${wrepo}/snapshots'/*/config.json"; then
  ok "weights already synced to worker (${wcount} files)"
```

Meanwhile `verify_snapshot()` — which does a real per-file byte check against
the HF manifest — is called **only from `download_on_head()`**. The worker never
runs it, in either `DOWNLOAD_MODE`.

## The concrete failure mode

A **truncated shard keeps the file count intact**. A killed rsync, a full disk
on the worker, an interrupted `snapshot_download` — the file exists, the name is
right, the count matches, the check passes. You find out at weight load, or you
don't find out at all and serve garbage.

The early return makes it durable: once a bad copy is in place, every subsequent
`./start.sh` re-accepts it without ever looking at the bytes.

Note this is only about the *worker*. The head is already covered.

## Why bytes are not redundant with the count

On the pair I develop against, the same pinned revision materialises as
**419 files / 135,253,622,894 bytes** in one cache and 421 files with a
different byte total in another (extra local files that are not in the HF
manifest). So the count is not even a stable identity for a revision — which is
exactly why I compare byte totals only where they are supposed to be
byte-identical (`rsync` mode, where the worker is a mirror of the head's repo
dir) and lean on `verify_snapshot`'s per-file manifest check everywhere else.

## The change

- **Byte total as well as count**, `find -L … -printf '%s\n'` so symlinks are
  dereferenced (a snapshot is a directory of links into `../../blobs/`; the
  link's own size is meaningless). `awk … printf "%.0f"`, not `%d` — mawk, which
  is `/usr/bin/awk` on Ubuntu, overflows a 135 GB total with `%d`.
- **`verify_snapshot` now runs on the worker too.** The python program moved
  into a `VERIFY_SNAPSHOT_PY` string so either node can run it; the head path is
  unchanged. It is shipped to the worker base64-encoded **as part of the
  command, never on stdin** — `wrun`'s ssh has to leave stdin alone or it eats a
  caller's heredoc (same hazard as #13). If the worker has no `python3` it warns
  and the byte-total compare still stands; if the HF API is unreachable the
  program's existing heuristic fallback applies, unchanged.
- Wired into **both** `DOWNLOAD_MODE`s and into the early returns.
- A count match with a **byte** mismatch `warn`s and falls through to the rsync
  (it resumes) instead of failing; anything still wrong after the sync fails
  closed, matching the surrounding style.

Not included: a `refs/main` byte-comparison. `resolve_snapshot()` already reads
it with `tr -d ' \n'`, which is more tolerant than the hub's own `f.read()`, so
there is nothing to fix there.

## Validation

`bash -n start.sh` clean. Then, on a live 2× GB10 pair (head + CX7-linked
worker, 419-file / 135.25 GB NVFP4 checkpoint, `HF_REVISION` sha-pinned), the
new helpers driven directly:

**Positive — the real trees:**

```
== head files/bytes ==
files: 419
bytes: 135253622894
== worker files/bytes ==
files: 419
bytes: 135253622894
== worker snapshot dir ==
/home/…/hub/models--RadixArk--Qwen3.8-Flash-Next-NVFP4/snapshots/7b71922524…
== verify_worker_weights ==
[OK]    worker byte total matches head (135253622894 bytes)
[OK]    worker cache: all 419 files verified (135.25 GB)
RESULT: PASS
```

That second `[OK]` is `verify_snapshot`'s own program, executed **on the
worker**, against the HF manifest.

**Negative — one shard truncated to 1 MiB** (a symlink mirror of the real
snapshot with a single `layer-00000-experts-0000-0127.safetensors` replaced by
its first 1 MiB, so the file count is untouched):

```
GOOD  files=419  bytes=135253622894
TRUNC files=419  bytes=134900573262      <-- count identical, 353 MB missing
```

The count check cannot see that. Both new checks can, independently:

```
== A. byte-total compare vs the real head total ==
[ERROR] worker bytes 134900571627 != head 135253622894 — truncated or partial shard; rerun to resume
RESULT: rejected (correct)

== B. same fixture, byte compare skipped — worker-side per-file check only ==
[ERROR] worker cache: size mismatch layer-00000-experts-0000-0127.safetensors (1048576 != 354098208)
RESULT: rejected (correct)
```

and on the head path, unchanged code, same fixture:

```
[ERROR] truncated fixture: size mismatch layer-00000-experts-0000-0127.safetensors (1048576 != 354098208)
RESULT: rejected, rc=1
```

Fixtures were symlink mirrors sharing the real blobs (~1 MiB of actual disk) and
were removed from both nodes afterwards. No serving process was restarted or
reconfigured for this.

I left `CHANGELOG.md` alone deliberately — I have two other small PRs open
against this file's neighbours and did not want three branches all editing the
same changelog block.

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -1548,9 +1548,9 @@ resolve_snapshot() { # → absolute host path of the snapshot (or fail)
   return 1
 }
 
-verify_snapshot() { # <snapshot_dir> <label> — byte-size check against the HF API
-  local dir="$1" label="$2"
-  python3 - "${MODEL_ID}" "${dir}" "${label}" <<'PY' || return 1
+# The per-file byte check, kept as a program string so it can run on either
+# node: python3 on the head, and shipped to the worker by verify_snapshot_worker.
+VERIFY_SNAPSHOT_PY="$(cat <<'PY'
 import json, os, sys, urllib.request
 model, snap, label = sys.argv[1], sys.argv[2], sys.argv[3]
 url = f"https://huggingface.co/api/models/{model}?blobs=true"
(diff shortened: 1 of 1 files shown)


MiaAI-Lab:
Reviewed and merging.

The count-only worker check is a genuine silent failure: a truncated shard keeps the file count, `config.json` is still there, and the early return re-accepts it on every later `./start.sh`. Head already had `verify_snapshot`; the worker did not.

Byte total (`find -L`, `awk %.0f` for mawk) plus shipping the per-file HF-manifest check to the worker as argv/base64 (not stdin) is the right pair of checks. Falling through to rsync on a count-match / byte-mismatch, then failing closed after sync, matches the surrounding style. Negative control in the PR body (1 MiB truncated shard, count identical) is convincing.

Head path is unchanged. Fine that CHANGELOG was left for a follow-up given the overlapping PRs.

PR #20, opened 2026-08-29
fix(weights): verify indexed shards offline

shuuul:
## Summary

- use `model.safetensors.index.json` as the offline source of truth
- verify every indexed shard exists and is non-empty
- avoid a redundant 135 GB download when the Hugging Face API is temporarily unavailable

The current fallback requires at least 400 safetensors, but this checkpoint has 206 indexed shards, so a complete local snapshot is rejected during an API outage.

## Verification

- `bash -n start.sh`
- forced the verifier's API request to a closed localhost port
- complete 206-shard snapshot passes
- synthetic snapshot with one missing indexed shard fails
- live two-node startup accepted both cached snapshots during an actual Hugging Face SSL failure

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,9 @@ All notable changes to this serving recipe.
 
 ### Fixed
 
+- Offline weight verification now reads the safetensors index instead of
+  assuming at least 400 shards; the model has 206 shards, so a transient HF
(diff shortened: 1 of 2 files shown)


shuuul:
Closing for now; we are pausing all upstream work while profiling and validating the local recipe.

PR #27, opened 2026-09-02
start.sh: honor the documented env knobs on both ranks + JSON override hooks

chuck-ads:
First off — thank you for this recipe. We run it in production on a dgx6+dgx2 pair and it's been rock solid.

## The bug

The launch heredocs hardcode `--enable-expert-parallel`, the MTP speculative config, and the mode-0 compilation config on both ranks, so several env knobs the script defines are silently ignored: `ENABLE_EXPERT_PARALLEL=false`, `MTP_NUM_SPECULATIVE_TOKENS=0`, `EXTRA_VLLM_ARGS`, and `EXTRA_DOCKER_ARGS` all no-op. We only noticed when an A/B we thought we were running turned out to be two identical launches.

## The fix

Both ranks now build identical engine-arg fragments from the env (resolved once, logged at launch so you can see what actually applied). Defaults preserve current behavior exactly — a stock `.env` produces the same `vllm serve` line as before.

Two new optional knobs, both defaulting to current behavior:
- `SPEC_CONFIG_JSON` — full `--speculative-config` JSON, for fields beyond `num_speculative_tokens` (e.g. `index_share_for_mtp_iteration`)
- `COMPILATION_CONFIG_JSON` — full `--compilation-config` JSON

One gotcha baked into the patch: the compile default uses an if-guard rather than `${VAR:-{json}}`, because bash ends that expansion at the first unescaped `}` and ships a trailing brace to `vllm serve` (we hit this in the wild).

## Why you might care

With these knobs working we ran a one-variable-at-a-time sweep on our pair and got single-stream decode from ~59 to **68.5 tok/s** (counting bench, temp 1.0) on top of your recipe — MTP `num_speculative_tokens: 4` + `index_share_for_mtp_iteration: true` + `--async-scheduling` (via `EXTRA_VLLM_ARGS`), plus disabling CPU C-states. Full numbers in the companion issue.

Tested on: 2× DGX Spark (GB10), TP2 over RoCE, native 262k, the stock container image. Happy to adjust style/naming to whatever you prefer.

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -64,6 +64,10 @@ KV_CACHE_DTYPE="${KV_CACHE_DTYPE:-auto}"   # auto = model dtype (bf16); this che
 PLE_OFFLOAD="${PLE_OFFLOAD:-false}"
 EXTRA_VLLM_ARGS="${EXTRA_VLLM_ARGS:-}"
 EXTRA_DOCKER_ARGS="${EXTRA_DOCKER_ARGS:-}"
+# Optional JSON overrides: full --speculative-config JSON (empty = MTP with
+# MTP_NUM_SPECULATIVE_TOKENS, 0 = off) and the --compilation-config JSON.
+SPEC_CONFIG_JSON="${SPEC_CONFIG_JSON:-}"
+if [[ -z "${COMPILATION_CONFIG_JSON:-}" ]]; then COMPILATION_CONFIG_JSON='{"mode":0,"cudagraph_mode":"FULL_DECODE_ONLY"}'; fi
 HF_TOKEN="${HF_TOKEN:-}"
 
 # YaRN only makes sense ABOVE the native 262144 context. At or below native,
@@ -473,6 +477,19 @@ else:
     PLE_OFFLOAD_ENV=""
     [[ "$PLE_OFFLOAD" == "true" ]] && PLE_OFFLOAD_ENV="-e VLLM_PLE_CPU_OFFLOAD=1"
 
+    # Resolve the engine-arg fragments shared by BOTH ranks (TP ranks must match).
+    EP_ARGS=""
+    [[ "$ENABLE_EXPERT_PARALLEL" == "true" ]] && EP_ARGS="--enable-expert-parallel --all2all-backend allgather_reducescatter"
+    if [[ -n "$SPEC_CONFIG_JSON" ]]; then
+        SPEC_ARGS="--speculative-config '$SPEC_CONFIG_JSON'"
+    elif [[ "$MTP_NUM_SPECULATIVE_TOKENS" -gt 0 ]]; then
+        SPEC_ARGS="--speculative-config '{\"method\":\"mtp\",\"num_speculative_tokens\":$MTP_NUM_SPECULATIVE_TOKENS}'"
+    else
+        SPEC_ARGS=""
+    fi
+    COMPILE_ARGS="--compilation-config '$COMPILATION_CONFIG_JSON'"
+    info "  engine fragments: EP=[$EP_ARGS] SPEC=[$SPEC_ARGS] COMPILE=[$COMPILE_ARGS] EXTRA=[$EXTRA_VLLM_ARGS] DOCKER_EXTRA=[$EXTRA_DOCKER_ARGS]"
+
     # Write worker launch script to a temp file and scp it (avoids SSH JSON quoting issues)
     WORKER_SCRIPT=$(mktemp /tmp/vllm_worker_XXXXXX.sh)
     cat > "$WORKER_SCRIPT" <<LAUNCH_EOF
@@ -482,6 +499,7 @@ docker run \
     --gpus all --network host --ipc host \
     --cap-add SYS_NICE --ulimit memlock=-1 --ulimit stack=67108864 \
     --device /dev/infiniband:/dev/infiniband \
+    $EXTRA_DOCKER_ARGS \
     -e GLOO_SOCKET_IFNAME=$WORKER_IFACE \
     -e NCCL_SOCKET_IFNAME=$WORKER_IFACE \
     -e TP_SOCKET_IFNAME=$WORKER_IFACE \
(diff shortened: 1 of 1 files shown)


sethforprivacy:
Independent confirmation of the `EXTRA_*` half of this, from a separate 2× GB10 pair (TP2 + EP + MTP3, image `sha256:d464f3b4…`, vLLM `v0.1.dev20073+g8e685d198`): we hit the same dead-code bug at `169fbad` and had patched it locally before finding this PR. Two extra data points in favor of merging:

- README:90 documents `EXTRA_VLLM_ARGS` as "appended last", and the 1M-context recipe at README:175 (`EXTRA_VLLM_ARGS="--kv-cache-memory=45502283776"`) depends on it — so the flag is not just unused today, a documented recipe silently no-ops.
- We inject auth as `--api-key` through `EXTRA_VLLM_ARGS`, so for us this was the difference between a keyed and a keyless listener on `0.0.0.0`. Worth noting the arrays `VLLM_ARGS`/`DOCKER_ARGS` built at `:394-396`/`:437-439` have no reader at all at `169fbad`.

One consequence this PR does not cover: once `EXTRA_VLLM_ARGS` is honored it can carry secrets, and the generated launch scripts are `chmod +x` (umask-dependent) under `/tmp`, with the worker copy left at the fixed path `/tmp/vllm_worker_launch.sh` after it runs. We put the 0700 + post-run removal in #29 so it is independent of this change and the two should not conflict.

jvr0x (changes requested):
@chuck-ads Requesting changes, and asking you to reduce scope. The premise of this PR is now about 75% obsolete, but one real bug survives and your PR is the right vehicle for it.

## The premise has moved

This was written against an old base where both `VLLM_ARGS` and `DOCKER_ARGS` were built and never read. Main has since fixed the `VLLM_ARGS` half: built at `start.sh:594-665`, rendered to `VLLM_ARGS_STR` at line 668, injected into both launch heredocs at lines 824 and 874. So `ENABLE_EXPERT_PARALLEL=false`, `MTP_NUM_SPECULATIVE_TOKENS=0` and `EXTRA_VLLM_ARGS` are all honored already, and `README.md:71-73` documents that. Three of your four knobs are fixed. The PR also currently conflicts with main.

## The bug that survives

`DOCKER_ARGS` at `start.sh:672-715` is still built and never read. There is no `DOCKER_ARGS_STR` and no `docker run "${DOCKER_ARGS[@]}"`; both heredocs hardcode every docker flag. Two consequences:

- `EXTRA_DOCKER_ARGS` (declared line 78, appended lines 712-714) is a no-op.
- `HF_TOKEN` (appended lines 709-711) never reaches either container. Nobody has reported this half.

This is a bug, not a design choice. The single-node variants both do it correctly: `start-tp1.sh:344,353` and `tp1/start.sh:343,352` inject `${HF_TOKEN:+-e HF_TOKEN=$HF_TOKEN}` and `$EXTRA_DOCKER_ARGS` straight into their heredocs. `README.md:155` documents `EXTRA_DOCKER_ARGS` as a working escape hatch.

## The ask

Rebase onto current main and reduce the PR to that one fix: inject `$EXTRA_DOCKER_ARGS` and `${HF_TOKEN:+-e HF_TOKEN=$HF_TOKEN}` into both launch heredocs, matching `start-tp1.sh`. If you would rather not rebase, say so and I will take the 5-line fix and credit you in the commit.

Credit where due: your handling of an empty `$EXTRA_DOCKER_ARGS` inside the heredoc is safe. I verified it collapses cleanly with no stray token, and `bash -n` passes.

## Please drop the two new knobs for now

`SPEC_CONFIG_JSON` and `COMPILATION_CONFIG_JSON`:

- `SPEC_CONFIG_JSON` bypasses two guards in main. `start.sh:407` errors when `MTP_DRAFT_VOCAB` is set with `MTP_NUM_SPECULATIVE_TOKENS=0`, and `start.sh:488` runs the MTP routed-expert quant-algo preflight. Both are keyed on `MTP_NUM_SPECULATIVE_TOKENS`. Setting `SPEC_CONFIG_JSON` while leaving that at 0 turns speculative decoding on and skips both guards, which reintroduces the silent unquantized-MoE failure the preflight exists to prevent. That is the exact bug just fixed in #39.
- Your `SPEC_ARGS` has no `MTP_DRAFT_VOCAB` branch, so it would drop `use_local_argmax_reduction` from the speculative config, which main adds at line 625.
- #45 is open right now and reworks the MTP and patch-stacking area, so these knobs would need redoing. Better to revisit after #45 lands.

## One correction for the record

Your commit message says the compile default uses an if-guard "because bash ends that expansion at the first unescaped `}` and ships a trailing brace to vllm serve". That diagnosis is wrong, and I tested it. On this bash, `${V:-{"mode":0}}` expands to `{mode:0}`, and `${V:-{"a":{"b":1}}}` expands to `{a:{b:1}}`. Bash matches the nested braces correctly and does not truncate. The real failure is that the unescaped inner double quotes are consumed as shell quoting, which strips the JSON's double quotes and produces invalid JSON. Your if-guard is still the correct fix, just for a different reason, and it is worth getting the reason right before it becomes repo lore.

## Two minor notes

Neither new knob is added to `.env.sample` or the README env table, so they are undiscoverable. And your new `info "  engine fragments: ... EXTRA=[$EXTRA_VLLM_ARGS] ..."` line prints `EXTRA_VLLM_ARGS` verbatim, which is how @sethforprivacy injects `--api-key`, so it logs a secret to stdout. `start.sh:742` on main already does the same, so it is pre-existing, but this widens it. Suggest redacting in both places.

chuck-ads:
@jvr0x Thanks for the thorough review. Going with option 1: I'll rebase onto current main and cut the PR down to the one surviving fix, injecting $EXTRA_DOCKER_ARGS and ${HF_TOKEN:+-e HF_TOKEN=$HF_TOKEN} into both launch heredocs to match start-tp1.sh.

Dropping SPEC_CONFIG_JSON and COMPILATION_CONFIG_JSON from this PR. I'll revisit those after #45 lands, with the guards keyed properly.

Agreed on the commit message. I'll reword it to say the inner double quotes get consumed by shell quoting rather than the brace-truncation claim. I'll also redact EXTRA_VLLM_ARGS in the info line I touch.

PR #29, opened 2026-09-02
fix: worker launch hygiene (exit status + no script left on the worker); document chat-template kwargs

sethforprivacy:
Two fixes found while bringing this recipe up on a 2× GB10 pair (TP2 + EP + MTP3, image `sha256:d464f3b4…`, vLLM `v0.1.dev20073+g8e685d198`) and then running it in production.

**Revised after @jvr0x's review** — the original description justified this PR partly with two README mismatches that are already fixed on main. That was wrong and is corrected below; the diff never contained a change for either. The review also caught a real regression in my own hardening change (the worker exit status), which is fixed.

**Note on scope:** we independently hit the `EXTRA_VLLM_ARGS` / `EXTRA_DOCKER_ARGS` dead-code bug. #27 already fixes that as a superset of our local patch, so this PR deliberately does **not** touch the arg expansion — please take #27 for it. What is left here is the consequence #27 does not cover, plus one doc fix.

### 1. `fix`: worker launch hygiene in `start.sh`

Two things, both in the generated-launch-script path:

**`chmod +x` is umask-dependent.** On a fresh `mktemp` file (0600), `chmod +x` yields 0711 under umask 022, 002 and 000 — 0700 only under 077. So a stock run leaves a world-readable launch script on both nodes under the common default umask. Harmless today, but once `EXTRA_VLLM_ARGS` is honored (#27) and carries `--api-key <key>` — which is how we inject auth on this lane — those scripts contain secrets.

The fix is to **remove** both `chmod` calls rather than tighten them: `mktemp` already creates 0600, the `+x` was the only thing loosening it, and both scripts are invoked as `bash <file>`, which runs fine on mode 0600. The exec bit was never needed. (An earlier revision of this PR used `chmod 700`; 0600 is strictly better, per review.)

The one place an explicit mode is still needed is the `.last_head_launch.sh` inspection copy: `cp` onto an existing file keeps *that file's* mode, so any tree that ever ran the `chmod +x` version stays 0711 there. Verified. `chmod 600` added.

**The worker copy is left behind at a fixed path.** `/tmp/vllm_worker_launch.sh` is predictable, survives the run, and — per review — `scp` without `-p` does not re-chmod an existing destination, so a mode set on the head side never reliably reaches the worker anyway.

Instead of scp'ing the file and cleaning it up afterwards, the script is now fed to the worker on **stdin**:

```sh
ssh_worker "bash -s" < "$WORKER_SCRIPT"
```

Nothing is written to the worker at all — no rendered `--api-key` at rest, nothing to clean up, no predictable `/tmp` name to pre-create as a symlink, and no dependence on `scp`'s chmod behaviour.

This also avoids the bug review caught in my first attempt at the cleanup. `ssh_worker "bash SCRIPT; rm -f SCRIPT"` returns the status of `rm -f`, so a worker `docker run` failure was reported as success; under `set -euo pipefail` the head then launched against a worker that never started, and the bare `while true` health loop turned a fail-fast into an infinite hang. Feeding on stdin makes ssh report the remote status directly. An explicit check names the failing code rather than aborting bare:

```
[ERR ]  Worker container failed to start (exit 3) — not launching the head.
```

Verified: worker exit 3 → `start.sh` exits non-zero with the head unlaunched and the temp file cleaned up (previously: exit 0, head launched); exit 0 still proceeds.

### 2. `docs`: chat-template kwargs

`reasoning_effort` accepts `low | medium | xhigh` (default `xhigh`) and 400s otherwise with `Unexpected reasoning effort max. Supported types are xhigh (default), medium, and low.` Both `high` and `max` are rejected, which surprises any client config shared with models where they are valid — one production request of ours hit exactly this.

Added `enable_thinking` alongside it, per review: default `true`, `false` prefills an empty `<think></think>` block. Plus one interaction worth documenting — the `reasoning_effort` validation sits *inside* the thinking branch, so with `enable_thinking: false` an invalid `reasoning_effort` is silently ignored instead of rejected. The same client config 400s in thinking mode and passes in non-thinking mode.

Placement fixed per review: this was misfiled under `## QSA launch profiles` and restated two facts the Multimodal callout already makes ~60 lines earlier (thinking consumes `max_tokens`; `thinking_budget` is not honored). Both restatements dropped; what is left is its own section immediately after `## Multimodal`, with a link back. Own section rather than a `###` under Multimodal because the kwargs are not multimodal-scoped — happy to fold it in if you prefer.

### 3. `docs`: page-cache note

Kept, but the justification in my original description was wrong. Main's "What Happens" step 5 is the MXFP8 patch and claims nothing about page caches; the only mention is the Gotchas bullet, already phrased as manual advice. What this adds is that it applies to **both** nodes and that `start.sh` does not do it for you (it needs no root — no `sudo` anywhere in `start.sh` or `files/nfs-share.sh`), stated where a first-time launcher reads it.

### Withdrawn

`--no-launch` — my description claimed the `start.sh` header, `--help` and the README Flags table disagreed about whether `--no-launch` prepares the PLE patch. On current main all three already agree, and this PR contains no change for it.

### Verification

- `bash -n start.sh` clean; `shellcheck -S warning start.sh` clean (remaining `info`-level findings are pre-existing and outside these hunks).
- `chmod +x` / `mktemp` / `cp` modes and the worker exit-status behaviour all reproduced locally before and after.
- `reasoning_effort`, `enable_thinking` and `thinking_budget` verified by rendering the checkpoint's `chat_template.jinja` directly.
- Field context: the recipe serves 1M context for us with every NIAH depth green to 976,125 real prompt tokens (3 depths per rung, first over-cap request 400s cleanly at admission with the engine healthy), KV pool 34.77 GiB / 2,455,108 tokens at the shipped `GPU_MEMORY_UTILIZATION=0.835` — within 1% of your README figure. Thanks for the recipe, and for a genuinely careful review.


jvr0x (changes requested):
@sethforprivacy Requesting changes. Two of the items in your description are already fixed on main, but the PR also contains two verified catches and one blocker.

## Blocker: the cleanup swallows the worker exit status

```
ssh_worker "bash /tmp/vllm_worker_launch.sh; rm -f /tmp/vllm_worker_launch.sh"
```

sshd runs the remote side as a non-interactive shell with no `set -e`. The exit status of the whole command is the status of `rm -f`, which is 0. Verified by simulation: main's form returns 3 when the worker script fails, your form returns 0.

Consequence: a worker `docker run` failure (bad image, GPU busy, invalid argument) is silently swallowed. The head then launches rank 0 against a worker that never started, and the health-check loop in `start.sh` is a bare `while true` that only breaks when the head container dies or `/health` returns 200. A fail-fast in seconds becomes an infinite hang.

```
ssh_worker "bash /tmp/vllm_worker_launch.sh; rc=\$?; rm -f /tmp/vllm_worker_launch.sh; exit \$rc"
```

## Major: `chmod 700` does not reach the worker

Your stated goal is mode 0700 on both nodes. The `chmod 700` applies only to the head-side temp file, which is `rm -f`'d immediately after the `scp`. `scp` is called without `-p`, and the destination is the fixed path `/tmp/vllm_worker_launch.sh`. If that file already exists, for example left at 0711 by a run before this PR, or by a run interrupted before the new `rm`, then `scp` does not re-chmod it. The worker copy can stay world-readable.

Fix: an explicit `ssh_worker "chmod 700 /tmp/vllm_worker_launch.sh"` after the scp, or better, scp to a randomised remote name. Also, neither your PR nor main addresses the symlink-attack surface of that fixed predictable path.

## Correct, and verified: `chmod +x` is umask-dependent

GNU coreutils 9.4, `chmod +x` on a fresh `mktemp` (0600) file: 0711 under umask 022, 0711 under 002, 0711 under 000, and 0700 only under umask 077. Main really does leave a world-readable launch script under the common default umask. Good catch.

Minor follow-on: the `+x` bit is unnecessary in the first place. The script is invoked as `bash /tmp/vllm_worker_launch.sh`, which I confirmed runs fine on a mode-0600 file. `chmod 600` is strictly better than `chmod 700`.

## Correct, and verified: `reasoning_effort`

Checked against the live `chat_template.jinja` from the checkpoint. Line 47 `reasoning_effort|default('xhigh')`, line 48 rejects anything not in `('xhigh', 'medium', 'low')`, and line 49 raises the exact error string you quote, character for character. `thinking_budget` really is absent from the template.

Optional: the live template also supports an `enable_thinking` kwarg (lines 46 and 165), documented nowhere in the README. Since you are already documenting template kwargs, it would be a good addition.

## Nit: placement

Your new block lands inside `## QSA launch profiles`, just before `## The PLE Patch`, not next to the endpoint docs as your description says. The endpoint material is the `## Multimodal` section, and `README.md:468-472` already states both facts you repeat there: thinking consumes `max_tokens`, and `chat_template_kwargs.thinking_budget` is not honored. The block duplicates them about 60 lines from the originals. Please move it next to them.

## Stale items in the description

Two of the three are already fixed on main, and your diff correctly contains no change for them:

- `--no-launch` is already consistent in all three places: `README.md:57`, `start.sh:18`, `start.sh:145`. All three already say the patch step does not run.
- Main's "What Happens" step 5 is the MXFP8 patch, and nothing in it claims that `start.sh` drops page caches. `grep -c drop_caches start.sh` is 0. The only README mention is the Gotchas bullet at lines 715-716, already phrased as manual advice. Keep your added note: it says "both nodes" and says explicitly that `start.sh` does not do it, so it is a net improvement. Only the justification in the description no longer matches main.

sethforprivacy:
Thanks — the blocker is real, it was mine, and you were right to hold the PR for it. Pushed 5f2bbb5 and 368358d, plus a rewritten description. Point by point.

## Blocker: worker exit status — fixed, and it took the `scp` with it

Reproduced your simulation exactly: main returns 3, my form returns 0. Under `set -euo pipefail` that means `ok "Worker container started."` prints on a dead worker, the head launches against it, and your bare `while true` health loop converts a fail-fast into an indefinite hang. Straightforwardly worse than what it replaced.

I went with a different fix than your suggested `rc=$?; rm; exit $rc`, because feeding the script on stdin resolves the blocker and your Major in one move:

```sh
ssh_worker "bash -s" < "$WORKER_SCRIPT"
```

Nothing is written to the worker at all. No rendered `--api-key` at rest, nothing to clean up, no fixed `/tmp/vllm_worker_launch.sh` to pre-create as a symlink, and no dependence on `scp`'s chmod semantics. `ssh` reports the remote status directly, so the `set -e` fail-fast is preserved, and I added an explicit check so the failure names itself instead of aborting bare:

```
[ERR ]  Worker container failed to start (exit 3) — not launching the head.
```

The worker script is `docker run -d`, so nothing depends on the remote stdin, and `/tmp/vllm_worker_launch.sh` was referenced nowhere else in the tree. Verified in situ against the patched block with a stubbed `ssh_worker`: exit 3 → start.sh exits non-zero, head unlaunched, temp file cleaned up; exit 0 → proceeds.

## Major: `chmod 700` never reached the worker — conceded

Correct, and the reasoning holds independently of the scp detail: I was chmod'ing a file that gets `rm -f`'d moments later and calling that "0700 on both nodes." With the file gone from the worker entirely, the question is moot.

Your point about a pre-existing destination keeping its old mode turned out to matter one line further down, too. `cp "$HEAD_SCRIPT" .last_head_launch.sh` onto an existing `.last_head_launch.sh` keeps *that file's* mode — I checked, 0711 in, 0711 out from a 0600 source. So on any tree that ever ran the `chmod +x` version, the inspection copy stays world-readable however tidy the source file is. The explicit `chmod 600` there is the one mode change in the diff that is actually load-bearing.

## `chmod +x` — and your follow-on is better than what I proposed

Reproduced on BSD `chmod` (macOS) as well as your GNU 9.4 numbers: `mktemp` gives 0600, then `+x` yields 0711 under umask 022, 002 and 000, and 0700 only under 077.

Taking your `chmod 600` note one step further: since `mktemp` is *already* 0600 and `+x` was the only thing loosening it, the right change is to delete both `chmod` lines rather than tighten them. Confirmed `bash <file>` runs a mode-0600 file. So the hardening is now a net **removal** of three lines from main's launch path rather than an addition, which is a better shape for this recipe than what I first sent.

## `reasoning_effort` — thanks, and I took the `enable_thinking` suggestion

Pulled the live `chat_template.jinja` and rendered it rather than take either of us at our word. Your line numbers are exact: 46, 47, 48, 49, and `enable_thinking` again at 165. `thinking_budget` count is 0, and a render with it is byte-identical to one without.

Documented `enable_thinking` as you suggested, and while rendering the matrix I hit an interaction worth adding: **the `reasoning_effort` validation lives inside the thinking branch** (line 46), so `enable_thinking: false` skips it entirely.

| kwargs | result |
|---|---|
| `reasoning_effort: high` / `max` | raises — HTTP 400 |
| `enable_thinking: false` + `reasoning_effort: max` | renders fine, silently ignored |

Same bad client config, 400 in thinking mode and a clean pass in non-thinking mode. That is in the new section.

## Nit: placement — fixed

Moved, and both restatements dropped; the section now links back to the Multimodal callout for those two facts instead of repeating them. One judgement call: I made it its own `## Chat-template kwargs` immediately after `## Multimodal` rather than a `###` inside it, since the kwargs are not multimodal-scoped. It is adjacent to the originals either way — say the word and I will fold it in as a subsection.

## Stale description items — my error, description rewritten

Both conceded; I checked them against current main rather than re-asserting them.

- `--no-launch`: consistent in all three places on main. Withdrawn, and called out as withdrawn.
- `drop_caches`: step 5 is the MXFP8 patch and claims nothing about page caches; the Gotchas bullet is already manual advice. I kept the note per your read, but the justification is rewritten to what it actually adds — "both nodes", and that `start.sh` does not do it (verified: no `sudo` anywhere in `start.sh` or `files/nfs-share.sh`).

Those two came from notes taken during our first cold launch and I carried them into the description without re-checking against a moving main. That is on me.

`bash -n` and `shellcheck -S warning` clean; remaining `info`-level shellcheck findings are pre-existing and outside these hunks. Still no change to the arg expansion — #27 remains the fix for that.

jvr0x:
Thanks for your contribution @sethforprivacy, merging.

PR #32, opened 2026-09-04
Uncensored mode with custom-quant PLE from orcarouter

jarvis959:
Hi Mia — thanks for the recipe, been running it on 2x DGX Spark.

This PR makes the PLE resolver override auto-derive instead of hardcoded `PLE_QUANT_OVERRIDE=fp8`, so a single `MODEL_ID` line in `.env` switches between compatible builds:

- `RadixArk/Qwen3.8-Flash-Next-NVFP4` (default, unchanged behavior) → `fp8`
- `lychee888/Qwen3.8-Flash-Next-Uncensored-NVFP4-FP8PLE` (orcarouter's abliterated build with the PLE n-gram table FP8-quantized in your PLE layout) → `fp8`
- `orcarouter/Qwen3.8-Flash-Next-Uncensored-NVFP4` (PLE kept bf16, 170.93 GiB, gated) → `bf16`

`start.sh` reads `text_config.ple_embedding_dtype` from the model's `config.json` (HF API, using `HF_TOKEN` when set; `float8_e4m3fn` → fp8, else bf16) unless `PLE_QUANT_OVERRIDE` is pinned in `.env`. All three builds share identical MTP weights and `layer_types`, so the MTP3 speculative config applies unchanged.

Verified on a live 2-node DGX Spark pair with the pinned image (`vllm/vllm-openai:qwen38-flash-next`):
- auto-derive: RadixArk config → `fp8` (matches today's hardcoded behavior), orca bf16 config → `bf16`, local FP8-PLE dir → `fp8`
- `bash -n` clean; `ple_layer_patched.py` shim untouched
- README gains a 'Recipe switching' section + `.env` table rows; `.env.sample` documents the switch

Happy to split or rebase if you'd prefer a smaller change. The FP8-PLE checkpoint docs (quantization details, RadixArk similarity) are at the [model card](https://huggingface.co/lychee888/Qwen3.8-Flash-Next-Uncensored-NVFP4-FP8PLE).

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -21,8 +21,25 @@ IB_HCA="=rocep1s0f0"                   # Single device, exact match (leading =)
 IB_GID_INDEX=3                          # Usually 3 on ConnectX with RoCE
 
 # -- Model ---------------------------------------------------------------------
+# THE RECIPE SWITCH — one line selects the checkpoint. Compatible builds:
+#   "RadixArk/Qwen3.8-Flash-Next-NVFP4"   (default) — the build this recipe targets
+#   "lychee888/Qwen3.8-Flash-Next-Uncensored-NVFP4-FP8PLE"
+#       — orca's abliterated build with the PLE n-gram table FP8-quantized
+#         (RadixArk-compatible PLE layout; 123.25 GiB; public).
+#   "orcarouter/Qwen3.8-Flash-Next-Uncensored-NVFP4"
+#       — orca's abliterated build, PLE kept bf16 (170.93 GiB; GATED: request
+#         access at huggingface.co and set HF_TOKEN below).
+#   "/absolute/path/to/model/dir" — local flat repo dir staged on both nodes
+#       (skips HF download + worker rsync; bind-mounted read-only). The path
+#       must resolve on both nodes (shared/NFS or identical staging).
 MODEL_ID="RadixArk/Qwen3.8-Flash-Next-NVFP4"
-SERVED_MODEL_NAME="qwen3.8-flash-next"
+SERVED_MODEL_NAME="qwen3.8-flash-next"          # for the orca-derived builds use "qwen3.8-flash-next-uncensored"
+
+# -- PLE quant override (optional) ---------------------------------------------
+# start.sh auto-derives this from the model's config.json
+# (text_config.ple_embedding_dtype: float8_e4m3fn -> fp8, absent/bf16 -> bf16).
+# Set explicitly only to override the auto-detection.
+# PLE_QUANT_OVERRIDE="fp8"
 
 # -- Serving Parameters --------------------------------------------------------
 MAX_MODEL_LEN=1000000                    # YaRN 1M context
@@ -62,6 +79,6 @@ PLE_OFFLOAD=false                       # Set true to offload 51B PLE table to C
 # -- Advanced ------------------------------------------------------------------
 # EXTRA_VLLM_ARGS=""
 # EXTRA_DOCKER_ARGS=""
-# HF_TOKEN=""
+# HF_TOKEN=""                            # needed only for gated repos (e.g. the orcarouter builds)
 VLLM_ALLOW_LONG_MAX_MODEL_LEN=1         # Required when MAX_MODEL_LEN > 262144 (YaRN)
 MASTER_PORT=50000                       # NCCL/distributed coordination port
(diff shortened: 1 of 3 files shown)


jvr0x:
@jarvis959 Closing this. The core mechanism the PR depends on no longer exists. There is a path forward at the bottom if you want to try again.

## The variable this PR sets does not exist

`PLE_QUANT_OVERRIDE` appears nowhere in main's executable code. Grepping the whole repo finds it only in prose: `README.md:318`, `README.md:410`, and `docs/CLAUDE/fable5-1-report.md:423`. Neither `files/ple_layer_patched.py` nor `files/patch_ple_layer.py` reads it. Every line this PR changes injects a dead environment variable, so the whole PR is a no-op on current main. It also conflicts.

Main already solved the underlying problem, and solved it better. It uses `PLE_EMBEDDING_DTYPE`, resolved by `files/detect_ple_dtype.py` reading the local cached `config.json` (`start.sh:241-247`), then fed to the engine through `--hf-overrides` (`start.sh:640-659`). No network call at launch time.

## Blocker: the README is corrupted

On your branch tip the literal text reads:

```
All numbers in **KV cache budget** and **Default runtime** below were rea## Recipe switching
```

and about 20 lines later:

```
d from the running
```

The word "read" has been split and the whole new section was spliced into the middle of the sentence. Because `## Recipe switching` is not at the start of a line, Markdown renders it as literal text rather than a heading, so the `[Recipe switching](#recipe-switching)` link that the same PR adds points at an anchor that does not exist.

## The detection logic is weaker than what it replaces

`files/detect_ple_dtype.py` documents three shapes that exist in the wild: `local-inference-lab/...` declares `nvfp4`, `RadixArk/...` declares `float8_e4m3fn`, and `nvidia/...` declares nothing and can only be recovered from `quantization_config.config_groups`. Confirmed against the live API: nvidia's `text_config.ple_embedding_dtype` is null while `config_groups.group_2` has `weights.num_bits = 8` targeting the PLE ngram embedding, which means FP8. Your PR reads only `text_config.ple_embedding_dtype`, so it returns `bf16` for the nvidia checkpoint, which is wrong, and it collapses `nvfp4` to `bf16`, also wrong. Main raises `SystemExit` on an unsupported `num_bits`; this PR swallows every error with `except Exception: print('')`.

It also fails open. I tested nine failure scenarios of the curl-based detection: DNS failure, HF 500, 429 rate limit, 401 on a gated repo with no token, the `--max-time 15` timeout, an HTML error page, `text_config: null`, and the key being absent. Eight of the nine silently produce `bf16` when the correct answer is `fp8`, and the only signal is one INFO line. This repo deliberately fails fast elsewhere: `start.sh:488-496` refuses to launch rather than build an unquantized MoE. This inverts that.

It also reads `resolve/main`, not the revision actually on disk, so the derived dtype can disagree with the weights in the local cache, and with no revision pin the bytes can change underneath a user.

## The documented local-path mode does not exist

`.env.sample` in this PR claims `"/absolute/path/to/model/dir"` works and "skips HF download + worker rsync; bind-mounted read-only". Tested: `MODEL_ID=/mnt/models/X` gives `ORG=""` and `NAME="X"` through the existing parameter expansion, so `MODEL_DIR` becomes `$HUB_PATH/models----X`, which does not exist, and `start.sh:216` aborts with "Could not resolve local cache path". Nothing in the diff implements skip-download, skip-rsync, or a bind mount.

Two minor notes: `HF_TOKEN` now appears in two rows of the same `.env` table with different descriptions, because the PR adds a row while the existing `EXTRA_VLLM_ARGS / EXTRA_DOCKER_ARGS / HF_TOKEN` row stays. And "a single MODEL_ID line switches" is not true: `SERVED_MODEL_NAME` must change too or clients get a 404 model-not-found, and nothing derives or validates it.

## On the submission history

Stating this neutrally: this branch is byte-identical to closed PR #31. `git diff pr-31..pr-32` is empty. #31 closed at 2026-09-04T05:50:11Z and this one opened at 05:51:09Z, 58 seconds later, with no changes and no closing comment. None of the above was addressed between submissions.

## Maintainer policy items

These are calls for the maintainer, not something a contribution can settle on its own. This PR rewrites the project's own README header to say "forked by jarvis959", and it promotes third-party "abliterated" re-quantizations from the official README of this repo, with no revision pin and no checksum, in a repo that is concurrently adding SHA-256 verification in #34.

## Path forward

If you still want recipe switching, the right base is main's existing `PLE_EMBEDDING_DTYPE` plus `detect_ple_dtype.py` mechanism, extended to read the local cache for whichever checkpoint is configured, and it must fail loudly rather than open. A reworked PR against that mechanism is welcome.

PR #39, opened 2026-09-06
Accept FP8_PB_WO as an alias of FP8_BLOCK_SCALES for MTP routed experts (fixes #38, part of #37)

### LINKED ISSUE #37: Quick start cannot complete: MTP experts declared FP8_PB_WO after 09-05 hub re-export; repo matches FP8_BLOCK_SCALES only
Environment

  - Hardware / nodes: 2× DGX Spark (GB10, 128 GB unified memory each); head + worker over ConnectX
  - Interconnect: ConnectX RoCE/IB (link-local 169.254.x.x; IFACE=enp1s0f0np0, IB_HCA="=rocep1s0f0,roceP2p1s0f0", IB_GID_INDEX=3)
  - Image / vLLM version: vllm/vllm-openai:qwen38-flash-next (docker image d464f3b466fa, the day-0 image from .env.sample; present on both nodes)
  - Model (MODEL_ID): nvidia/Qwen3.8-Flash-Next-NVFP4 — the .env.sample default. Local snapshot is hub main (fc694b54…); config.json / hf_quant_config.json
    verified sha256-identical to resolve/main.
  - start.sh flags used: ./start.sh --no-download (cache populated first; Step 3 confirmed the worker already has the weights)
  - Relevant .env values: MAX_MODEL_LEN=262144 (YARN_ENABLE=false), KV_CACHE_DTYPE=fp8, GPU_MEMORY_UTILIZATION=0.835, TENSOR_PARALLEL_SIZE=2,
    ENABLE_EXPERT_PARALLEL=true, MTP_NUM_SPECULATIVE_TOKENS=3, MAX_NUM_SEQS=8, MAX_NUM_BATCHED_TOKENS=8192, PLE_OFFLOAD=false, FP8_DENSE=false,
    QSA_PROFILE=stock, PORT=8888
  - Repo state: upstream main c2325b2 (checked at filing time; no fix pushed yet)

  ---

  Steps to Reproduce

  Following the README Quick Start on a fresh clone:

  1. cp .env.sample .env, fill in IPs/interfaces (only network values edited — model/MTP/cache settings left at .env.sample defaults)
  2. ./download.sh (or populate the HF cache with the current hub revision any way; cache verified complete: all 11 safetensors + config files, configs
     sha256-identical to resolve/main)
  3. ./start.sh --no-download
  4. Watch the launch abort at Step 4g before any container starts

  Expected results: the launch proceeds through the patch steps and starts the vLLM container with the README's default MTP3 configuration (as documented in
  the "Checkpoint: nvidia/Qwen3.8-Flash-Next-NVFP4" and "KV cache budget" sections).

  Actual results: abort in start.sh at Step 4g with MTP experts are FP8_PB_WO …, exit 1. The MTP layer-index alias step itself succeeds first (MTP experts
  alias added to: config.json hf_quant_config.json) — the failure is the subsequent preflight. Setting MTP_NUM_SPECULATIVE_TOKENS=0 is the only
  error-suggested workaround and loses the MTP3 speedup.

  ---

  Additional context

  Root cause (evidence-backed):

  1. nvidia/Qwen3.8-Flash-Next-NVFP4 main was re-exported on 2026-09-05 (fc694b54, "Fix MTP serving metadata and instructions"). Diffing against the 09-03
     revision (fab0aecb): one line of config.json changed — the MTP routed experts' quant_algo FP8_BLOCK_SCALES → FP8_PB_WO. All weight shards,
     model-fp8-mtp-ple.safetensors and hf_quant_config.json are byte-identical (git blob oids verified).
  2. Same layout, renamed string: both declare group_size: 128; the checkpoint stores weight_scale_inv per 128-block; the model card says the MTP routed
     experts are 128×128 block-scaled FP8 "copied byte-for-byte" from Qwen/Qwen3.8-Flash-Next-FP8. The image itself knows FP8_PB_WO (QUANT_ALGOS in
     modelopt.py, with a dense-layer method) — it is ModelOpt's canonical name for per-block weight-only FP8.
  3. The recipe matches the old name only: files/patch_modelopt_fp8_block_moe.py adds the RoutedExperts dispatch branch for quant_algo == "FP8_BLOCK_SCALES",
     and files/patch_checkpoint_config.py SUPPORTED_MOE_ALGOS = {FP8, NVFP4, W4A16_NVFP4, MXFP8, FP8_BLOCK_SCALES}. The Step 4g preflight (--mtp-moe-algo)
     reads config.json first — the one file the rename touched — misses the allowlist and aborts.
  4. The guarded hazard is real: ModelOptMixedPrecisionConfig.get_quant_method returns None for unknown RoutedExperts algos → silently unquantized MoE →
     AttributeError: … has no parameter 'w2_weight_scale_inv' deep into the load. But note vLLM feeds the per-layer dispatch dict from hf_quant_config.json
     (get_config_filenames() in weight_utils.py), which the rename did not touch (still FP8_BLOCK_SCALES in both revisions) — so the dispatch may actually
     build the MTP MoE fine from the legacy sidecar, and the preflight's config.json-first read trips on a name the dispatch path never consumes.

  Suggested fix (small, in existing files): in files/patch_modelopt_fp8_block_moe.py accept both names (if quant_algo in ("FP8_BLOCK_SCALES", "FP8_PB_WO"): —
  same builder, block shape already read from group_size), and add "FP8_PB_WO" to SUPPORTED_MOE_ALGOS in files/patch_checkpoint_config.py.

  Secondary robustness finding: start.sh:239 resolves the config snapshot with ls -d snapshots/*/ | head -1; with ≥2 snapshot dirs in the HF cache (e.g. an
  interrupted download leaves a partial revision dir) it can pick an incomplete dir and Step 4g dies with patch_checkpoint_config: no config.json at ….
  Preferring the revision refs/main names would fix that.

  <details>
  <summary>Minimal reproduction sample</summary>

  No API sample — the server is never reached: start.sh exits before docker run.

  </details>

  <details>
    <summary>Logs</summary>

  Output from start.sh (the run aborted here, exit 1):

  $ ./start.sh --no-download
  [INFO]  === Step 2: Resolve cache path ===
  [ OK ]  Model cache: /home/liuya/.cache/huggingface/hub/models--nvidia--Qwen3.8-Flash-Next-NVFP4
  [ OK ]  PLE table dtype not declared by checkpoint — overriding to float8_e4m3fn
  [INFO]  Head HF cache:   /home/liuya/.cache/huggingface
  [INFO]  Worker HF cache: /home/liuya/.cache/huggingface
  [INFO]  === Step 3: Sync weights to worker (169.254.246.185) ===
  [ OK ]  Worker already has models--nvidia--Qwen3.8-Flash-Next-NVFP4 — skipping rsync.
  [INFO]    (delete /home/liuya/.cache/huggingface/hub/models--nvidia--Qwen3.8-Flash-Next-NVFP4 on the worker to force a re-sync)
  [INFO]  === Step 4: Verify weights on head ===
  [ OK ]  HEAD  (169.254.221.195): /home/liuya/.cache/huggingface/hub/models--nvidia--Qwen3.8-Flash-Next-NVFP4 (136G)
  [INFO]  === Step 4b: GPU preflight ===
  [ OK ]  Both GPUs idle.
  [INFO]  === Step 4f: FP8 KV cache patch (fp8) ===
  patched qsa_ops_patched.py
  patched qsa_nvidia_patched.py
  ok
  [WARN]  FP8 KV is a quality trade on sparse attention - validate reasoning on your workload.
  [INFO]  === Step 4g: MTP layer-index alias ===
  [ OK ]  MTP experts alias added to: config.json hf_quant_config.json
  [ERR ]  MTP experts are FP8_PB_WO, which this image's mixed-precision MoE
         dispatch cannot build (supports FP8 / NVFP4 / W4A16_NVFP4 / MXFP8).
         Set MTP_NUM_SPECULATIVE_TOKENS=0 in .env to serve without speculative
         decoding, or use a checkpoint whose MTP experts are NVFP4.

  The log is from start.sh — no docker logs vllm-fn exists because the container was never launched.

  </details>

### LINKED ISSUE #38: NVIDIA checkpoint MTP preflight fails on FP8_PB_WO despite hf_quant_config.json reporting FP8_BLOCK_SCALES
## Summary

The current dual-DGX recipe fails during the MTP preflight when serving:

`nvidia/Qwen3.8-Flash-Next-NVFP4`

with the current NVIDIA checkpoint snapshot:

`fc694b54fb0174e0913e6adf86691ef85a4ead47`

Repository revision tested:

`c2325b2`

The launcher aborts before model load with:

    [INFO]  === Step 4g: MTP layer-index alias ===
    [ OK ]  MTP experts alias added to: config.json hf_quant_config.json

    [ERR ]  MTP experts are FP8_PB_WO, which this image's mixed-precision MoE
             dispatch cannot build (supports FP8 / NVFP4 / W4A16_NVFP4 / MXFP8).
             Set MTP_NUM_SPECULATIVE_TOKENS=0 in .env to serve without speculative
             decoding, or use a checkpoint whose MTP experts are NVFP4.

## Root cause

The two quantization metadata files in the current NVIDIA checkpoint describe the same MTP routed expert using different algorithm names.

`config.json` contains:

    "mtp.layers.0.mlp.experts": {
      "quant_algo": "FP8_PB_WO",
      "group_size": 128
    }

while `hf_quant_config.json` contains:

    "mtp.layers.0.mlp.experts": {
      "quant_algo": "FP8_BLOCK_SCALES",
      "group_size": 128
    }

`files/patch_checkpoint_config.py::mtp_moe_algo()` currently checks:

    for name in ("config.json", "hf_quant_config.json"):

and immediately returns the first matching MTP expert algorithm.

Therefore the preflight sees `FP8_PB_WO` from `config.json` and exits with status 3 before considering the legacy sidecar.

This is particularly significant because the recipe already contains explicit runtime support for `FP8_BLOCK_SCALES` through:

`files/patch_modelopt_fp8_block_moe.py`

The checkpoint tensors also contain the expected block-FP8 scale tensors, for example:

    mtp.layers.0.mlp.experts.0.down_proj.weight_scale_inv
    mtp.layers.0.mlp.experts.0.gate_proj.weight_scale_inv
    mtp.layers.0.mlp.experts.0.up_proj.weight_scale_inv

which matches the `Fp8MoEMethod` path used by the recipe.

## Workaround / proposed fix

Changing the preflight metadata precedence from:

    for name in ("config.json", "hf_quant_config.json"):

to:

    for name in ("hf_quant_config.json", "config.json"):

allows the preflight to resolve:

    FP8_BLOCK_SCALES

and the existing supported path is then selected:

    [ OK ] MTP experts quantization: FP8_BLOCK_SCALES (supported)

The launch then proceeds through:

- FP8 KV patch
- MTP layer-index aliases
- PLE patch
- MXFP8 fallback patch
- FP8_BLOCK_SCALES ModelOpt MoE patch
- worker launch
- head launch
- NCCL world size 2 initialization
- TP2 + EP initialization
- MTP3 initialization
- checkpoint loading

without changing the checkpoint files themselves.

## Why I preferred this workaround

I did not simply add `FP8_PB_WO` to `SUPPORTED_MOE_ALGOS`.

The existing runtime patch explicitly implements `FP8_BLOCK_SCALES`, while the legacy `hf_quant_config.json` already identifies this exact MTP expert as `FP8_BLOCK_SCALES`.

Preferring the sidecar therefore uses the metadata representation that matches the existing runtime implementation rather than assuming globally that `FP8_PB_WO` is always synonymous with `FP8_BLOCK_SCALES`.

## Environment

- 2 × NVIDIA DGX Spark
- TP=2
- Expert Parallel enabled
- MTP=3
- `nvidia/Qwen3.8-Flash-Next-NVFP4`
- checkpoint snapshot `fc694b54fb0174e0913e6adf86691ef85a4ead47`
- repo revision `c2325b2`
- `vllm/vllm-openai:qwen38-flash-next`
- `KV_CACHE_DTYPE=fp8`
- `MAX_MODEL_LEN=262144`
- `GPU_MEMORY_UTILIZATION=0.835`

## Current validation status

After applying the metadata-precedence workaround, the launch proceeds successfully past the original failure point.

Observed so far:

    MTP experts quantization: FP8_BLOCK_SCALES (supported)

    Resolved architecture: Qwen4ExpForConditionalGeneration

    Resolved architecture: Qwen3_8FlashNextMTP

    world_size=2

    Expert parallelism is enabled.
    Local/global number of experts: 256/512

    Using 'FLASHINFER_CUTLASS' NvFp4 MoE backend

    PLE embedding ... uses the runtime FP8 method

The NVIDIA checkpoint then begins loading normally:

    Checkpoint size: 123.57 GiB

    Loading safetensors checkpoint shards:
      0% Completed | 0/11
      9% Completed | 1/11
     18% Completed | 2/11
     27% Completed | 3/11

I will update this issue with final `/health` confirmation once the full model load completes.

Happy to provide the complete startup log if useful.

wojo:
## Description and Motivation

`nvidia/Qwen3.8-Flash-Next-NVFP4` was re-exported on 2026-09-05 (hub commit `fc694b54`, "Fix MTP serving metadata and instructions"). That commit changed exactly one line: the MTP routed experts' `quant_algo` in `config.json` went from `FP8_BLOCK_SCALES` to `FP8_PB_WO`. `hf_quant_config.json` still says `FP8_BLOCK_SCALES` for the same layer, and every weight shard is byte-identical (`group_size: 128`, per-128-block `weight_scale_inv`). `FP8_PB_WO` is ModelOpt's canonical name for per-block weight-only FP8; the image's `modelopt.py` already lists it, but only with a linear-layer method.

The recipe matches the old string exactly in two places, so a fresh clone against the current hub snapshot fails at Step 4g:

```
[ERR ]  MTP experts are FP8_PB_WO, which this image's mixed-precision MoE
        dispatch cannot build (supports FP8 / NVFP4 / W4A16_NVFP4 / MXFP8).
```

This PR treats `FP8_PB_WO` as an alias of `FP8_BLOCK_SCALES` in both places, so both names take the path this recipe already validated:

- `files/patch_checkpoint_config.py`: `SUPPORTED_MOE_ALGOS` gains `FP8_PB_WO` (with a comment recording the upstream rename).
- `files/patch_modelopt_fp8_block_moe.py`: the routed-expert branch becomes `if quant_algo in ("FP8_BLOCK_SCALES", "FP8_PB_WO"):` and routes to the same `Fp8MoEMethod` with `weight_block_size` read from `group_size`. The log line now prints the actual algo instead of a hardcoded `FP8_BLOCK_SCALES`.
- The idempotency guard keyed on `"FP8_BLOCK_SCALES" in src`. That would skip a `modelopt_patched.py` produced by the previous version of the script (which already contains the old `==` branch) and never add the alias. It now keys on the new branch text — the stock `modelopt.py` already contains `FP8_PB_WO` for linear layers, so that string alone is not a usable marker — and upgrades an old-form file in place.

Both halves are required: as #38 found, widening only the preflight lets the launch proceed and then fail ~7 min into loading with `Layer mtp.layers.48.mlp.experts has no parameter 'w2_weight_scale_inv'` because the runtime dispatch falls through to `None`.

No `.env` knob changes. No effect on measured numbers — the weights and block shape are unchanged; this restores the MTP3 configuration the README documents. Safe on both nodes: the patched `modelopt_patched.py` is the same file `start.sh` already scp's to the worker.

## Related Issues

Fixes #38.

Partially addresses #37 — this resolves its primary finding (the `FP8_PB_WO`
preflight abort and the matching runtime dispatch gap). It deliberately does
**not** close #37: that issue carries a second, still-unfixed finding
(`start.sh:239` resolving the snapshot with `ls -d snapshots/*/ | head -1`, which
can pick a partial revision directory), noted as out of scope at the bottom of
this description. #37 stays open to track it.

#37 contains the hub-revision diff establishing that only the label changed; #38 contains the runtime failure that shows the preflight-only workaround is insufficient.

---

## Testing

- `python3 -m py_compile` on both files.
- Preflight, against a copy of the current hub `config.json` (`FP8_PB_WO`): shipped script → `FP8_PB_WO  exit=3`; patched script → `FP8_PB_WO  exit=0`.
- Runtime patch, against the real `modelopt.py` from `vllm/vllm-openai:qwen38-flash-next`, three paths: fresh (after `patch_modelopt_mxfp8.py`) → `ok`, branch present at the dispatch site, file byte-compiles; second run → `already patched`; a `modelopt_patched.py` generated by the previous script version (old `==` branch) → `upgraded (FP8_PB_WO alias)`, one new-form branch, byte-compiles.
- Actual launch on 2× ASUS Ascent GX10 (GB10), TP2+EP over ConnectX-7 200 GbE, checkpoint `fc694b54`, MTP3, `KV_CACHE_DTYPE=fp8`, `GPU_MEMORY_UTILIZATION=0.835`:

```
Model loading took 64.3 GiB memory
Available KV cache memory: 35.31 GiB
GPU KV cache size: 3,784,544 tokens, Maximum concurrency for 262,144 tokens per request: 14.44x
```

`bench/mtp_accept.py`: 64.6% acceptance, 1.94 accepted/draft, positions 81.8% / 64.3% / 47.6%. Decode 47–53 tok/s single-stream vs 26–27 with `MTP_NUM_SPECULATIVE_TOKENS=0`.

Not in this PR (separate concern from #37): `start.sh` resolving the snapshot with `ls -d snapshots/*/ | head -1`, which can pick a partial revision directory.

https://claude.ai/code/session_01QcJBXab6sCmZji1r9ugsbK

### DIFF
--- a/files/patch_checkpoint_config.py
+++ b/files/patch_checkpoint_config.py
@@ -81,10 +81,13 @@ def add_aliases(quant_config: dict, num_hidden_layers: int) -> bool:
 # RoutedExperts algos ModelOptMixedPrecisionConfig.get_quant_method can build.
 # Anything else resolves to a *silently unquantized* MoE, which then dies at load
 # with "has no parameter 'w2_weight_scale_inv'".
+# FP8_PB_WO is ModelOpt's newer name for the same 128x128 block-scaled FP8 weights
+# (nvidia/Qwen3.8-Flash-Next-NVFP4 commit fc694b54, 2026-09-05, renamed it in
+# config.json only; hf_quant_config.json still says FP8_BLOCK_SCALES).
 # FP8_BLOCK_SCALES is supported only because files/patch_modelopt_fp8_block_moe.py
 # adds that branch; stock vLLM (image and upstream main) would build an
 # unquantized MoE and die at load.
-SUPPORTED_MOE_ALGOS = {"FP8", "NVFP4", "W4A16_NVFP4", "MXFP8", "FP8_BLOCK_SCALES"}
+SUPPORTED_MOE_ALGOS = {"FP8", "NVFP4", "W4A16_NVFP4", "MXFP8", "FP8_BLOCK_SCALES", "FP8_PB_WO"}
 
 
 def mtp_moe_algo(snapshot_dir: str) -> str:
--- a/files/patch_modelopt_fp8_block_moe.py
+++ b/files/patch_modelopt_fp8_block_moe.py
@@ -80,15 +80,16 @@ DISPATCH = """            if quant_algo == "MXFP8":
                     quant_config=self.mxfp8_config,
                     moe_config=layer.moe_config,
                 )
-            if quant_algo == "FP8_BLOCK_SCALES":
+            if quant_algo in ("FP8_BLOCK_SCALES", "FP8_PB_WO"):
                 # Imported lazily: modelopt.py deliberately does not import fp8.py
                 # at module scope.
                 from vllm.model_executor.layers.quantization.fp8 import Fp8MoEMethod
 
                 logger.info_once(
-                    "Routed experts %s use FP8_BLOCK_SCALES; building them with "
+                    "Routed experts %s use %s; building them with "
                     "Fp8MoEMethod (block-quantized).",
                     prefix,
+                    quant_algo,
                 )
                 return Fp8MoEMethod(
                     quant_config=self._fp8_block_scales_config(prefix),
@@ -104,10 +105,21 @@ def _replace_once(src: str, old: str, new: str, what: str) -> str:
     return src.replace(old, new)
 
 
+NEW_BRANCH = 'if quant_algo in ("FP8_BLOCK_SCALES", "FP8_PB_WO"):'
+OLD_BRANCH = 'if quant_algo == "FP8_BLOCK_SCALES":'
+
+
 def patch() -> None:
     src = open(TARGET).read()
-    if "FP8_BLOCK_SCALES" in src:
+    if NEW_BRANCH in src:
         print("already patched", TARGET)
+        return
+    if OLD_BRANCH in src:
+        # Patched by an earlier version of this script: widen the branch to the
+        # FP8_PB_WO alias without re-applying the helper.
+        src = _replace_once(src, OLD_BRANCH, NEW_BRANCH, "moe dispatch (alias upgrade)")
+        open(TARGET, "w").write(src)
+        print("upgraded (FP8_PB_WO alias)", TARGET)
         return
     src = _replace_once(src, ANCHOR_HELPER, HELPER + ANCHOR_HELPER, "helper")
     src = _replace_once(src, ANCHOR_DISPATCH, DISPATCH, "moe dispatch")


jvr0x:
Merging this. Thanks for you contribution @wojo! one bookkeeping change, the verification I ran before merging, and two follow-ups that are **not** blockers.

### #37 is not being closed by this PR

I edited the description (`Fixes #37. Fixes #38.` → `Fixes #38.` + a partial-address note) and the title, so the merge doesn't auto-close #37. Flagging that openly since it's your PR text I changed.

Reason: #37 carries two findings. This PR fixes the first one completely - the `FP8_PB_WO` preflight abort and the matching runtime dispatch gap. The second is the one you flagged yourself as out of scope at the bottom of your description:

> Not in this PR (separate concern from #37): `start.sh` resolving the snapshot with `ls -d snapshots/*/ | head -1`, which can pick a partial revision directory.

That is still unfixed - `start.sh:239` is unchanged on main - and no open PR touches snapshot resolution. Auto-closing #37 would bury it, so it stays open to track that item.

### What I verified before merging

I confirmed the central claim against the live hub rather than taking it on faith:

- `nvidia/Qwen3.8-Flash-Next-NVFP4` HEAD is `fc694b54` (2026-09-05, "Fix MTP serving metadata and instructions"), as you said.
- `config.json` → `mtp.layers.0.mlp.experts: {quant_algo: FP8_PB_WO, group_size: 128}`
- `hf_quant_config.json` → same layer `{quant_algo: FP8_BLOCK_SCALES, group_size: 128}`
- `num_hidden_layers: 48`, so the alias target is `mtp.layers.48`.

`group_size: 128` being identical in both files is what makes routing both names to the same `Fp8MoEMethod` safe, and that holds.

I also extracted `modelopt.py` from the pinned image (`vllm/vllm-openai:qwen38-flash-next`, `d464f3b4`) to test the idempotency-marker reasoning, since that is the non-obvious part of this change:

- stock contains `FP8_BLOCK_SCALES` **0** times
- stock contains `FP8_PB_WO` **8** times - `QUANT_ALGOS` at l.105, a `quant_method ==` branch at l.389, the linear method class at l.620+

So your note is exactly right: `FP8_PB_WO` alone is unusable as a marker, and keying on the full branch text is the correct call. It is also more robust than the guard it replaces. `FP8_PB_WO` is already in the image's `QUANT_ALGOS`, so if upstream ever ships its own `FP8_BLOCK_SCALES` support, the old `"FP8_BLOCK_SCALES" in src` test would false-positive as "already patched" and silently skip installing both the dispatch branch and `_fp8_block_scales_config`. The duplicate PR #40 (closed) left the old guard in place, so this is the better of the two fixes.

One thing worth recording for anyone reading #38 later: the alternative proposed there - reversing `mtp_moe_algo` precedence to read `hf_quant_config.json` first - is not merely insufficient, it is wrong in principle. @drmarcopapa's own follow-up on #38 proves the runtime dispatch receives `FP8_PB_WO`, i.e. `config.json` is authoritative for `quantized_layers`. That independently confirms your "both halves are required", and it means the existing `config.json`-first order in `mtp_moe_algo` should stay as it is.

### Two follow-ups - merging without them, no action needed from you

1. **The upgrade path leaves the old log statement behind.** `_replace_once(src, OLD_BRANCH, NEW_BRANCH, ...)` replaces only the `if` line, so a file upgraded in place still logs a hardcoded `Routed experts %s use FP8_BLOCK_SCALES` when the algo is actually `FP8_PB_WO` - a misleading diagnostic in precisely the scenario this PR exists for. In practice it is unreachable through `start.sh`: `files/modelopt_patched.py` is gitignored and `patch_modelopt_mxfp8.py` rebuilds it from `.orig` on every launch, so this only affects a standalone re-run against a stale file. The branch itself is justified - without it `ANCHOR_DISPATCH` counts 0 in an old-form file and `_replace_once` raises `AssertionError`. It just needs the log fix carried across too.

2. **`start.sh:468-470` and `patch_checkpoint_config.py:main()` both assert that "vLLM reads the legacy file when present".** The runtime evidence on #38 contradicts that. Patching both files is still the right defensive behaviour; the stated *reason* is wrong, and someone trusting it could drop the `config.json` mount and break MTP.

I'll take both in a follow-up commit alongside the `refs/main` snapshot fix from #37, so you don't need to do anything unless you'd rather own them.

### Why this jumped the queue

`.env.sample` ships `MODEL_ID="nvidia/Qwen3.8-Flash-Next-NVFP4"` with `MTP_NUM_SPECULATIVE_TOKENS=3`, so a fresh clone following the README aborts at Step 4g today - that is #37 exactly as filed. `RadixArk/…` is unaffected (its `config.json` declares no `quantized_layers` at all, so `mtp_moe_algo` short-circuits and the guard never fires), which is presumably why this went unnoticed since the 09-05 re-export. Two independent cluster validations, both halves of the fix, and the reasoning behind the marker change spelled out - this was the most thoroughly evidenced of the PRs currently open. Thanks.

PR #44, opened 2026-09-07
perf: FP8-dense + quantized MTP + 32k balanced draft vocab (+44% mean, +51% prose)

KamakuraCrypto:
## What this does

Takes single-stream decode on the 2-Spark kit from **36.7 → 55.3 tok/s prose**
(mean 44.3 → 64.0) and adds a single-Spark serving path.

`bench/decodebench.py`, 1k context, greedy, every run validated clean (exact
expected generation-token count, because a co-resident agent on this box queries
the server continuously and silently contaminates deltas):

| config | prose | code | entropy | copy | mean | acceptance |
|---|---|---|---|---|---|---|
| baseline (this repo) | 36.7 | 44.1 | 38.2 | 58.3 | 44.3 | 53.6% |
| + 32,768-id draft vocab | 44.6 | 49.1 | 45.3 | 58.9 | 49.5 | ~50% |
| + FP8-dense | 54.8 | 61.8 | 55.1 | 72.0 | 60.9 | 48.9–52.0% |
| **+ MTP layer quantized** | **55.3** | **62.6** | **59.1** | **79.1** | **64.0** | 47.0–49.1% |

That is **+30% prose / +41% code over this repo's previously published best**
(42.5 / 44.5 / 45.4 / 56.5, the `--balance-shards 2` row).

Concurrency, `max_num_seqs=48` (new `bench/concbench.py`):

| streams | 1 | 8 | 16 | 32 | 48 |
|---|---|---|---|---|---|
| aggregate tok/s | 54.6 | 197.7 | 295.2 | 356.7 | **448.3** |

## The four changes

**1. A 32,768-id corpus-ranked draft vocabulary.** The MTP drafter owns its own
248,320-row lm_head, read once per draft step — three of the four lm_head reads in
an MTP=3 engine step. Ranking over a 16.5M-token corpus yields 60,407 distinct
ids, so a 32,768 cut is *entirely* corpus-ranked at 99.2% coverage with no
`--fill-to-size` merge-order guessing. Coverage is the whole ballgame: an 8,600-id
cut tried first dropped acceptance 53.6% → 34.3% and cost 20% on code.

**2. FP8-dense serving together with the draft vocabulary.** These were mutually
exclusive — both overlay `nvidia/mtp.py`. They are textually disjoint, but there
is a real semantic collision: FP8-dense makes the drafter's `lm_head` an FP8
tensor with per-row scales while the draft-vocab patch slices it and calls
`F.linear`, which fails ~10 minutes into launch. `files/stack_mtp_fp8_draftvocab.py`
bridges it. The same pattern unblocks FP8-KV + the GB10 QSA profile on
`ops/qsa.py`, and `modelopt.py`.

**3. The MTP draft layer quantized** (`MTP_FP8=true files/fp8dense/build.sh`).
`mtp.*` was excluded from the converter, leaving the whole draft layer BF16 —
~0.59 GiB/step, re-read on every draft pass. CPU-only, 78 s, ~7.4 GiB of new bytes
(rest hard-linked); run per node rather than rsyncing ~120 GiB.

**4. Draft-head storage balancing** (`VLLM_MTP_DRAFT_VOCAB_BALANCE=1`, opt-in).
Byte-level BPE puts frequent tokens at low ids, so slicing by the lm_head's own
shard range put **32,406 of 32,768 rows on rank 0 and 362 on rank 1**; a decode
step waits for the slowest rank. Now 16,384 each. This is *not*
`build_draft_vocab.py --balance-shards`, which balances the *vocabulary* and cost
acceptance 56.5% → 47.8% — here the token set is unchanged and acceptance was
unchanged.

## Scope of the safety claim

The **drafter** changes are distribution-preserving: rejection sampling is exact
for any proposal `q`, so a truncated or quantized drafter can only lower
acceptance, never change what is served.

**FP8-dense is not in that category** and should not be described as output-safe.
It quantizes the *target* — attention, GDN projections, HyperConnection, target
`lm_head` — affecting every request, including ones with speculation disabled.
Measured weight error is rel RMSE 0.026 / cosine 0.9997, but that bounds *weight*
error, not generation quality. **Task-level validation is outstanding**; run the
GSM8K/AIME harness, and do not gate on perplexity (documented on this model
family: AIME 86.7 → 80.8 while long-context perplexity *improved*).

Distribution-preserving also is not seed-reproducible — changing proposals changes
accept/reject decisions and therefore RNG consumption.

## Bugs fixed along the way

- `QSA_PROFILE != stock` was **broken**: `apply_patch.py` reads `qsa.py.orig`,
  which nothing extracted and `.gitignore` hid. `FileNotFoundError` on any clean
  checkout.
- `files/overlay/apply_patches.py` **did not exist**; `start.sh:386` called it.
  Same `.gitignore` cause.
- `EXTRA_DOCKER_ARGS` **never reached either node** in the 2-node path: head and
  worker launch from heredocs with their own `-e` lists and `DOCKER_ARGS` builds
  neither. Added `VLLM_EXTRA_ENV`, spliced into both.
- The balancing path could **deadlock the cluster** (found by adversarial review,
  fixed in `a09af5e`): collectives sat inside a broad `try/except` that fell back
  *locally*, so one rank escaping would leave its peer blocked in `all_gather`
  forever. Now a single `all_reduce(MIN)` every rank reaches decides globally, the
  install is transactional, and postconditions assert the reassembled id set.

## Also included

- `autoresearch/` — config sweep harness with a hard launch deadline and automatic
  rollback (a `torch.compile` mode-3 experiment hung both nodes for 100 minutes
  before this existed), contamination detection, and `RESULTS.md` recording what
  worked, what failed, and what is **ruled out** so it is not re-run.
- `files/ple_nvfp4/` + `tp1/` — the PLE n-gram table is 40% of the checkpoint and
  was the last large FP8 component; at NVFP4 the checkpoint is 98.70 GiB and fits
  one Spark resident. Predicted ~40 tok/s prose (+60% over the published
  single-Spark 24.9). Honest caveat: **two independent single-Spark servers lose to
  one TP=2 pair** at every concurrency measured (−13% at 8 streams, −19% at 16,
  −9% at 48) because TP=2+EP splits the touched experts. That path is for
  isolation, not throughput. PLE at 4-bit also measures rel RMSE 0.094 — 3.6× the
  FP8-dense error — and RadixArk's published evals do not transfer to it.

## Negative results worth not repeating

`torch.compile` mode 3 hangs GB10 (Inductor ran 103 min, killed both nodes; it
also logs `Not enough SMs to use max_autotune_gemm mode` at 48 SMs). GPU clock
locking does nothing (the GPU self-limits at ~2470 MHz regardless; no memory-clock
control is exposed). CPU pinning to the X925 cores does nothing. `flashinfer_b12x`
MoE is EP-incompatible. `--all2all-backend` is inert at `dp_size=1`. The GB10 QSA
launch profile measured *worse* (−6.5%) despite plausible reasoning. Details and
evidence in `autoresearch/RESULTS.md`.


KamakuraCrypto:
Closing to rework: rewriting the branch with a tighter, scope-limited history.

PR #45, opened 2026-09-07
Extend FP8 quantization to the MTP layer, unblock stacked patches, add test tooling

KamakuraCrypto:
## What this is

Changes across four areas of the serving path, the tooling to test them, and three
bug fixes found on the way. **Output quality is not validated — see Caveats.**

Roughly where the code went:

| area | ~lines | |
|---|---|---|
| checkpoint quantization | 260 | extend the FP8-dense converter to the MTP draft layer |
| patch stacking | 410 | three stackers + the missing `apply_patches.py` |
| startup / config | 97 | new `.env` knobs, env that reaches both nodes, stacking orchestration |
| draft vocabulary | 128 | 32k list + TP storage balancing |
| test tooling | 470 | sweep harness + concurrency bench |

### Checkpoint quantization (`files/fp8dense/`)

The FP8-dense converter previously excluded `mtp.*` entirely, so the whole MTP
draft layer stayed BF16 and was re-read on every draft pass. Added:

- `--mtp-dense` — per-channel FP8 for the draft layer's attention projections and
  its three HyperConnections, using the same recipe as the main model.
- `--mtp-experts` — the 512 routed experts as 128x128 block-scaled FP8 E4M3 with
  `weight_scale_inv` multipliers, emitted per expert (the fused 3-D form has no
  scale-tensor mapping in the loader). This is the same `FP8_BLOCK_SCALES` format
  the `nvidia/Qwen3.8-Flash-Next-NVFP4` MTP already ships.
- `mtp_alias()` — the checkpoint names these `mtp.layers.0.*` but vLLM builds the
  layer at the absolute index `mtp.layers.<num_hidden_layers>` and looks
  `quantized_layers` up by that runtime prefix (`nvidia/mtp.py` remaps only
  `exclude_modules`), so the metadata now carries both spellings in both
  `config.json` and the legacy `hf_quant_config.json`.
- `--link-unchanged-from` — hard-link untouched shards from an earlier snapshot, so
  a rebuild costs ~7.4 GiB of new bytes rather than a full copy. CPU-only, ~78 s;
  cheap enough to run per node instead of rsyncing ~120 GiB.

Verifier and quant-stats scripts extended to cover the new tensors.

### Startup and configuration (`start.sh`)

- **`VLLM_EXTRA_ENV`** — arbitrary `KEY=VALUE` env spliced into *both* node
  heredocs. This exists because `EXTRA_DOCKER_ARGS` reached **neither** node in the
  2-node path: head and worker are launched from separate heredocs with their own
  `-e` lists, and `DOCKER_ARGS` builds neither. Env that changes kernel selection
  must reach both ranks, or they run different kernels and a decode step simply
  waits for the slower one.
- **`COMPILATION_MODE` / `CUDAGRAPH_MODE`** — previously hardcoded to
  `{"mode":0,"cudagraph_mode":"FULL_DECODE_ONLY"}` with no knob and no recorded
  rationale. Now configurable, defaults unchanged.
- **`MTP_DRAFT_SAMPLE_METHOD`** — exposes `draft_sample_method`, with a guard: it
  is mutually exclusive with `MTP_DRAFT_VOCAB` because the reduced draft head is
  only reachable through `use_local_argmax_reduction`, which `probabilistic`
  forbids. `start.sh` now refuses the combination loudly instead of silently
  dropping one.
- Orchestration for the stackers below, including conditional overlay mounting so
  the same container path is never bind-mounted twice.

### Patch stacking

`start.sh` previously *refused* combinations where two patch sets touch the same
vendor file. Three stackers apply one patch onto another patch's output:

- `stack_mtp_fp8_draftvocab.py` — FP8-dense + reduced draft vocabulary on
  `nvidia/mtp.py`. Textually disjoint, but they collide semantically: FP8-dense
  makes the drafter's `lm_head` an FP8 tensor with per-row scales while the
  draft-vocab patch slices that head and calls `F.linear`, failing ~10 minutes into
  launch. The stacker bridges it by dequantizing the slice with the matching scales.
- `stack_qsa_fp8_gb10.py` — FP8-KV + the GB10 QSA launch profile on `ops/qsa.py`.
  They collide on exactly one anchor (the MQA scorer launch); order matters.
- `stack_modelopt_fp8dense.py` — folds the FP8-dense hunks into the already-patched
  `modelopt.py`.

Each asserts its anchors and refuses to write a half-applied file.

### Draft vocabulary

- A 32,768-id list (`draft_vocab_32k.txt`) ranked over a 16.5M-token corpus, which
  yields 60,407 distinct ids — so the cut is entirely corpus-ranked with no
  `--fill-to-size` merge-order padding (the earlier 65,536 list padded ~87% of its
  slots that way). Coverage dominates: an 8,600-id list from a small corpus
  measured *worse* than the full vocabulary on code.
- **Opt-in TP storage balancing** (`VLLM_MTP_DRAFT_VOCAB_BALANCE=1`). Byte-level
  BPE puts frequent tokens at low ids, so slicing by the lm_head's own shard range
  put 32,406 of 32,768 rows on rank 0 and 362 on rank 1; a decode step waits for
  the slowest rank. Redistributes to 16,384 each. Distinct from
  `build_draft_vocab.py --balance-shards`, which balances the *vocabulary* and is
  documented here as costing acceptance 56.5% -> 47.8%; this leaves the token set
  untouched. The decision is made collectively and installed transactionally, so a
  one-sided failure cannot leave a peer blocked in `all_gather`. Off by default.

## Bugs fixed

- **`QSA_PROFILE != stock` could not work.** `files/qsa_gb10/apply_patch.py` reads
  `qsa.py.orig`, which nothing extracts and `.gitignore` hides —
  `FileNotFoundError` on any clean checkout.
- **`files/overlay/apply_patches.py` did not exist**, though `start.sh:386` calls
  it. Same `.gitignore` cause. This blocked `FP8_DENSE` entirely.
- **`EXTRA_DOCKER_ARGS` reached neither node** in the 2-node path: head and worker
  are launched from heredocs with their own `-e` lists, and `DOCKER_ARGS` builds
  neither. Added `VLLM_EXTRA_ENV`, spliced into both — important because the two
  ranks must run identical kernels.
- `COMPILATION_MODE` / `CUDAGRAPH_MODE` / `MTP_DRAFT_SAMPLE_METHOD` are now `.env`
  knobs rather than hardcoded, with a guard on the `draft_sample_method` vs
  `use_local_argmax_reduction` incompatibility.

## Testing tooling

**`autoresearch/`** — a sweep harness that applies `.env` overrides, relaunches,
benchmarks, records to `results.jsonl`, and restores the baseline. Three
properties that turned out to matter:

- **Hard launch deadline with rollback.** A `torch.compile` experiment hung both
  nodes for 100 minutes before this existed. A trial that misses its deadline is
  killed, recorded `FAILED` with its log tail, and the baseline is restored.
- **Contamination detection.** Any co-resident client polling the server that does
  not set `temperature` inherits the checkpoint's `generation_config.json` default
  (`temperature=1.0, top_k=20, top_p=0.95`), which drags acceptance down and reads
  as a regression that is not there. Each run checks its generation-token delta
  against the expected count and marks the run `DIRTY` if other traffic intruded.
- **Records which backend actually engaged**, because a flag that silently does
  nothing is easy to credit with a win — `--all2all-backend` is inert at
  `dp_size=1`, for example.

**`bench/concbench.py`** — aggregate-throughput sweep across concurrency levels,
reporting per-stream, aggregate, and engine steps/s.

## Caveats — please read before trusting any of this

**Output quality is not validated.** What was measured is *numerics*: weight-level
rel RMSE ~0.026 and cosine ~0.9997 for FP8-dense. That bounds weight error. It
does **not** bound generation quality, and no task-level evaluation was run.
Anyone adopting this should run the GSM8K/AIME harness and
`bench/reasoning_check.py` first. Note that perplexity is a deceptive gate on this
model family — a published result has AIME dropping 86.7 → 80.8 while long-context
perplexity *improves*.

**The changes are not equally safe.** The drafter-side changes (reduced vocabulary,
quantized MTP layer, storage balancing) only alter the proposal distribution, and
rejection sampling is exact for any proposal — they can lower acceptance but not
change what is served. **FP8-dense is different**: it quantizes the target model
(attention, GDN projections, HyperConnection, target `lm_head`) and affects every
request, including ones with speculation disabled. It should be treated as an
explicit quality mode, not a free win.

**Distribution-preserving is not seed-reproducible.** Changing which tokens are
proposed changes accept/reject decisions and therefore RNG consumption, so a fixed
seed can produce a different (valid) sequence.

**Known residual.** If one rank returns early from `_attach_draft_vocab`
(unreadable vocab file, unexpected FP8 scale shape) while its peer proceeds, the
peer can still block. It needs an agreement exchanged before the early returns.
Requires an asymmetric mount between nodes to trigger.

## On the numbers

`autoresearch/RESULTS.md` records measurements taken on one 2-Spark kit with
`bench/decodebench.py` at 1k context, greedy, runs validated for contamination.
They are offered as evidence that the mechanisms do something on that kit, **not
as portable benchmarks** — decode tok/s is dominated by draft acceptance, which is
a property of the prompt. Comparing tok/s across setups with different prompts
compares workloads, not engines; engine steps/s is the prompt-independent number,
and `bench/concbench.py` reports it. Reproduce on your own workload before
believing anything here.

`RESULTS.md` also records the **negative** results, which are arguably the more
useful part: `torch.compile` mode 3 hangs GB10; GPU clock locking and CPU pinning
do nothing measurable; `flashinfer_b12x` MoE is EP-incompatible; the GB10 QSA
launch profile measured worse despite plausible reasoning.


jvr0x (changes requested):
Requesting changes. The two bug fixes are real and verified, but the stacking mechanism has defects that must be fixed first.

## What is good, verified

`files/overlay/apply_patches.py` was missing while `start.sh:378` calls it, so FP8_DENSE died on a clean checkout. `qsa.py.orig` was never extracted, so any non-stock QSA profile died the same way. Both confirmed on main. Fixing them is genuine value.

The converter asserts BF16 source before quantizing, so a wrong source fails loud. The built checkpoint declares FP8_BLOCK_SCALES, which fits the merged #39 fix. The balance block is well built: agreement first, transactional install, postcondition checks. The sample-method guard is the right design.

I also ran the full modelopt sequence (mxfp8, fp8_block_moe, stack) against the stock `modelopt.py` from the pinned image. All layers survive, including the #39 alias branch. The stacker is compatible with #39.

## Must fix

1. The QSA stacker destroys `qsa.py.orig`. It copies FP8 output over the pristine file. Regeneration guards skip files that exist, so only manual deletion restores it. A later switch to bf16 plus GB10 then mounts the stale stacked file with no warning. Use a separate input path. Do not clobber `.orig`.
2. Five new env knobs with zero docs: `VLLM_EXTRA_ENV`, `COMPILATION_MODE`, `CUDAGRAPH_MODE`, `MTP_DRAFT_SAMPLE_METHOD`, `VLLM_MTP_DRAFT_VOCAB_BALANCE`. None is in README or `.env.sample`.
3. `VLLM_EXTRA_ENV` leaves the documented knobs dead. `EXTRA_DOCKER_ARGS` and `HF_TOKEN` still go to the dead `DOCKER_ARGS` array. Wire them or remove them. Also the comment is wrong: it says the old knob reaches only the head, but it reaches neither.
4. The new `info` line prints `VLLM_EXTRA_ENV` verbatim, and that knob is the designated secrets channel. Redact values or drop the line.
5. The admitted hang stays open. An early return in `_attach_draft_vocab` can block the peer forever. The description names the fix but does not build it.

## Smaller items

- The stacker rewrites the tracked `apply_patch.py` in place. Right now that branch is dead because the committed file already carries the change, but the design still dirties source on launch. Compute it in memory or use a temp copy.
- `files/fp8dense/start.sh.fp8dense-draftvocab.diff` only records the `start.sh` diff, which git already holds. Drop it or say what reads it.
- The committed 32k vocab cannot be regenerated or audited here since the corpus is not included. Note the provenance.
- The `autoresearch/` harness is large for this PR. Fine to keep, but the parts that relaunch and kill the cluster deserve their own review pass.

kesslerio:
Thanks for sharing this — I liked the approach here, especially separating full-vocabulary probabilistic drafting from reduced-vocabulary greedy drafting and validating the setup on both ranks. I tried the relevant pieces in my fork on my dual-GB10 TP2 setup to see how they behaved with my actual agent workload.

For context, I kept the RadixArk NVFP4 target, FP8 KV, 262K context, six active sequences, compilation mode 0 and stock QSA. My client normally runs at T=1, top_p=.95 and top_k=20.

The final comparison was my original shared-KV/batch8192 setup versus an explicit 24-billion-byte KV pool with batch4096. Both used full-vocabulary greedy MTP3. I ran three repetitions of 12 fixed repository, repair, tool and long-context tasks at one and three concurrent tasks:

- original setup: 68/72 raw checks passed
- KV24/batch4096: 71/72 passed
- at c1, median successful-task time improved 12.3%, while p90 was 4.3% slower
- at c3, median improved 28.7%, while p90 was 14.8% slower

I ended up keeping my original setup because the concurrent tail got noticeably worse, even though the median was much better. The smaller batch did help smooth active streaming during an incoming cold 128K prefill: the median longest gap dropped from 3.17s to 1.67s.

Short decode performance was much closer to the headline numbers. My original setup measured roughly 38.4–41.5 tok/s solo and about 142 tok/s aggregate at c6. The KV24/batch4096 warmup measured 41.0 tok/s solo code and 140.9 tok/s aggregate at c6. In my use, the bigger difference comes from long prefills, tool turns and completion-time tail rather than short solo decode.

I also tried a few related options. Full-vocabulary probabilistic MTP3 improved decode, but I hit an Ofus QSA/Triton worker fault during concurrent tool work. The balanced 32K greedy vocabulary improved some short decode results but missed my real-task threshold, and a separate target dense-FP8 trial did not pass my speed and regression checks.

My test harness and Qwen-specific changes are in [my fork PR #6](https://github.com/kesslerio/Qwen3.8-Flash-Next-Dual-DGX-Sparks/pull/6) if any of the details are useful. Thanks again — the approach and review comments gave me a much better set of things to test than just chasing the best single-stream tok/s number.

PR #53, opened 2026-09-14
Enable agentic prefix-cache observability and preserve local test configuration

mrexodia:
## Description and Motivation

This PR improves prefix-cache observability for agentic and tool-loop workloads.

### Launcher changes

`start.sh` now applies the following options through the shared `VLLM_ARGS` array, ensuring identical behavior on both TP ranks:

- `--enable-prefix-caching`
- `--enable-prompt-tokens-details`
- `--kv-cache-metrics`
- `--kv-cache-metrics-sample "$KV_CACHE_METRICS_SAMPLE"`

A new `.env.sample` knob is documented:

```bash
KV_CACHE_METRICS_SAMPLE=1.0
```

This enables:

- Exact-prefix reuse for repeated agent conversations
- `prompt_tokens_details.cached_tokens` in API usage responses
- Prefix-cache and KV residency metrics from `/metrics`
- Per-request cache measurements in sparkDash Pi replay benchmarks

vLLM continues to derive the safe prefix-match unit from the hybrid QSA/GDN cache geometry. With BF16 SSM state, live launches selected:

- MTP=1: 1,616-token pages
- MTP=3: 1,664-token pages

The PR intentionally does not force a smaller match unit.

### Test-safety fix

`tests/test_check_weights.sh` previously deleted an existing local `.env` when it exited. It now saves and restores the file, including its original permissions. This prevents running the test suite from destroying a configured cluster installation.

### Live measurements

Tested with the upstream-supported `RadixArk/Qwen3.8-Flash-Next-NVFP4` checkpoint on two DGX Sparks using:

- TP2 + EP
- MTP=3
- FP8 KV
- BF16 SSM state
- Native 262,144-token context
- `GPU_MEMORY_UTILIZATION=0.835`

The restored MTP=3 launch reported:

```text
Available KV cache memory: 34.58 GiB
GPU KV cache size: 4,182,588 tokens
Maximum concurrency for 262,144 tokens per request: 15.96x
```

A repeated 37,143-token Pi agent payload reached:

- 34,944 cached tokens
- 94% prefix-cache hit rate
- 1.01 s warm TTFT

Post-first-token decode is not expected to improve directly from prefix caching; the primary benefit is reduced repeated-prefill work and observable cache accounting.

The same validation also compared MTP=3 against MTP=1. MTP=3 remained the general-purpose choice:

| Workload | MTP=3 | MTP=1 |
|---|---:|---:|
| Prose C1 | 47.91 tok/s | 41.82 tok/s |
| Prose C2 aggregate | 82.09 tok/s | 74.03 tok/s |
| Prose C4 aggregate | 130.24 tok/s | 126.03 tok/s |
| Eight-workload mean | 40.53 tok/s | 39.61 tok/s |

## Related Issues

None.

---

## Testing

Completed:

- `bash -n start.sh stop.sh download.sh check-weights.sh`
- `bash tests/test_check_weights.sh` — 12/12 passed
- `python3 -m unittest tests/test_verify_weights.py` — 40/40 passed
- Verified that `tests/test_check_weights.sh` preserves the original `.env` byte-for-byte
- Actual dual-node launch with `./start.sh --launch`
- Confirmed `/health` and `/v1/models`
- Confirmed both ranks received the shared prefix-cache and metrics arguments
- Confirmed `prompt_tokens_details.cached_tokens` in live responses
- Pi agent payload replay with cache and speculative-acceptance deltas
- sparkDash prose benchmark at concurrency 1, 2, and 4
- `bench/reasoning_check.py` — 12/12, including a 104,088-token needle prompt

`shellcheck` was not available on the host.

---

## Checklist

- [x] I have read the README and `.env.sample` and kept my changes consistent with them.
- [x] My pull request has a sound title and description.
- [x] My change is reproducible and verified with an actual dual-node launch and re-measurement.
- [x] I updated the README and `.env.sample` for the new knob and runtime behavior.
- [x] Defaults in `.env.sample` still work out of the box, and `KV_CACHE_METRICS_SAMPLE` has a fallback of `1.0`.


deme4gaming-web:
Independent confirmation from a second live deployment (2× DGX Spark / GB10, recipe `d2ae28a`). The destructive path is real, and it fires on a *configured* install, not only in CI.

- We ran `bash tests/test_check_weights.sh` in our checkout; it printed `All 12 shell tests passed.`
- Afterwards the checkout's real `.env` (8212 B, a working TP2 cluster config) was gone.
- Cause: `check-weights.sh` does `cd "$SCRIPT_DIR"` and reads `.env` relative to its own directory (lines 32-33 and 68), while the test writes its fixture into the repo root (`cd "$REPO_ROOT"` line 16, `cat > .env` line 132) and the EXIT trap removed that same path (`rm -rf "$WORK" .env`, line 20). Nothing failed loudly - the suite reported success while deleting the deployment's config.
- We restored ours from a pre-update backup and verified it byte-identical (sha256 `d9e421f1...`).

Thanks for the save/restore here - it closes the normal path. Two gaps remain that a copy-based variant closes:

1. **Hard kill / timeout**: on SIGKILL (OOM killer, CI timeout, closed terminal) the `EXIT` trap never runs, so the *fixture* `.env` survives in the checkout. The previous failure mode left no `.env` at all (loud `ERROR: .env not found`); the new one leaves a plausible-looking fixture that a concurrent `./start.sh` would consume silently.
2. **Window during the run**: the fixture sits at the real repo root for the whole run, so any concurrent reader (`start.sh`, `stop.sh`, a human) sees test values.

Suggested shape - run the script under test from a throwaway copy, so the checkout's `.env` is never written or removed at all:

```diff
@@
 REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
-cd "$REPO_ROOT"
 
 WORK="$(mktemp -d)"
 export WORK
-trap '[[ -n "${SERVER_PID:-}" ]] && kill "$SERVER_PID" 2>/dev/null || true; rm -rf "$WORK" .env' EXIT
+trap '[[ -n "${SERVER_PID:-}" ]] && kill "$SERVER_PID" 2>/dev/null || true; rm -rf "$WORK"' EXIT
+
+# check-weights.sh resolves .env and its helpers relative to its OWN directory
+# (SCRIPT_DIR), so running it from the repo root would make this test write -
+# and delete - the checkout's real .env. Run it from a copy instead.
+SANDBOX="$WORK/repo"
+mkdir -p "$SANDBOX/files"
+cp "$REPO_ROOT/check-weights.sh" "$REPO_ROOT/verify-weights.py" "$SANDBOX/"
+cp -a "$REPO_ROOT/files/." "$SANDBOX/files/"
+cd "$SANDBOX"
```

Verified on that same box: with the original test the checkout's `.env` disappears; with the change above `All 12 shell tests passed` both in a copied tree and in the live checkout, with the real `.env` sha256 unchanged after both runs (1 file, +13/-2).

Happy to open a small PR in that shape if you would prefer it over save/restore, or to leave it as a suggestion here.

PR #60, opened 2026-09-16
Normalise refs/main whitespace so hand-staged caches resolve offline (fixes #36)

### LINKED ISSUE #36: Offline resolution fails for hand-staged caches: hf_hub 1.28 reads refs/main without strip()
The new `start.sh` passes `MODEL_ID` (repo id) to the container and relies on
`HF_HUB_OFFLINE=1` cache resolution. That works for hub-downloaded caches but
breaks for **hand-staged / NAS-staged** ones — a flow this recipe explicitly
supports (NFS staging, rsync-your-own-hub-dir, `--no-download`).

**Cause:** this image ships `huggingface_hub 1.28.0`. Its offline branch in
`_snapshot_download.py` (~line 312) reads the ref with `commit_hash = f.read()`
— **no `.strip()`**. Hand-staged `refs/main` files usually end with a newline
(`echo $SHA > refs/main`), so the computed snapshot path becomes
`snapshots/<sha>\n`, `os.path.exists()` fails, and the engine dies at arg-parse
with:

```
huggingface_hub.errors.LocalEntryNotFoundError: Cannot find an appropriate
cached snapshot folder for the specified revision on the local disk and
outgoing traffic has been disabled.
```

The snapshot folder itself is fine — only the trailing newline differs.

**Fix that worked:** rewrite refs without the newline on both nodes:

```bash
for r in ~/.cache/huggingface/hub/models--*/refs/main; do
    printf '%s' "$(cat "$r")" > "$r"
done
```

Suggest `start.sh` sanitize `refs/main` during Step 2/3 (cache resolution) —
`printf '%s'` instead of trusting the file byte-for-byte — since every
NAS-staging path in the README's ecosystem (`llamas-to-nas`, flat-dir → hub
conversion) is exposed to this.

jvr0x:
Fixes #36.

## The bug

This image ships `huggingface_hub 1.28.0`. Its offline branch reads the ref with a bare `f.read()` and **no `.strip()`**. A ref staged the obvious way carries a trailing newline:

```
$ echo "$SHA" > refs/main
$ od -c refs/main
0000000   a   b   c   1   2   3  \n
```

So offline resolution builds a path that cannot exist:

```
computed path: 'hub/models--Fake--Model/snapshots/abc123\n'
exists? False          <-- the bug
with .strip(): True
```

vLLM then dies at arg-parse with a repo-not-found error naming neither the file nor the newline.

`hf download` writes the ref clean, so this only bites the staging flows the README advertises: **NFS staging, rsync-your-own-hub-dir, `--no-download`**. I hit the setup for it while staging a checkpoint by hand earlier today and only avoided it by using `printf '%s'` on instinct.

## The fix

`start.sh` normalises the ref at cache-resolution time on the head, and on the worker in the rsync path. Under `NFS_SHARE` the worker mounts the head's cache read-only, so the head-side pass already covers it.

Rewrites **only when the bytes actually differ**, so hub-downloaded caches are untouched and silent.

One deliberate refusal: a ref that is *entirely* whitespace is left alone rather than blanked. An empty ref is a different failure and should surface as itself, not be quietly turned into a zero-byte file.

## Verification

**End-to-end, against the real cache.** Deliberately broke a staged `refs/main` with `echo`, then ran `start.sh`:

```
[WARN]  Normalised trailing whitespace in .../models--RadixArk--.../refs/main
        (hf_hub 1.28 offline resolution reads this file without .strip(); see issue #36)
[ OK ]  Model cache: .../models--RadixArk--Qwen3.8-Flash-Next-NVFP4
        (snapshot 7b719225242aacd3dbd3f9407468c2ee9a9d2594)
```

Immediate re-run: **0 warnings**. A different checkpoint's ref in the same cache: unmodified.

**Unit-checked** across every input shape:

| input | result |
|---|---|
| `echo`-written (`\n`) | normalised, resolves |
| already clean | no-op, no warning |
| CRLF (`\r\n`) | normalised, resolves |
| trailing spaces + `\n` | normalised, resolves |
| run 3× | idempotent |
| whitespace-only ref | **left alone**, not blanked |
| missing file / path | returns cleanly, no crash |

## Scope

One function plus one worker-side call, +34 lines, no behaviour change on a hub-downloaded cache.


jvr0x:
Pushed a second commit. Two things came out of testing this properly rather than declaring it done.

## 1. The worker-side check was broken

The original worker branch compared the stripped value against `$(cat "$ref")`. **Command substitution strips trailing newlines**, so both sides were always equal and the rewrite never fired. The check was structurally blind to the only byte it exists to detect.

My first end-to-end test had `NFS_SHARE=true`, so it took the NFS branch and never ran this code at all. Exercising the rsync path exposed it immediately: `start.sh` reached the code, printed nothing, and left the worker's ref ending in `\n`.

Now compares byte count to stripped length. Re-verified over ssh against the real worker: newline-terminated ref → warning printed, trailing byte removed; immediate re-run silent.

## 2. Confirmed the root cause against the real library

The original report describes `huggingface_hub 1.28.0` reading the ref without `.strip()`. I had only *simulated* that in Python. Tested it against the actual library in the actual image:

**Broken ref (`abc123\n`):**
```
hf_hub version: 1.28.0
FAILED -> LocalEntryNotFoundError
   Cannot find an appropriate cached snapshot folder for the specified revision
   on the local disk and outgoing traffic has been disabled...
```

**Same cache, newline stripped (`abc123`):**
```
RESOLVED -> '/probe/hub/models--Fake--Model/snapshots/abc123'
```

One byte, and the error names neither the file nor the newline — which is why this costs a launch to diagnose.

## Verification now covers

| layer | how |
|---|---|
| Root cause | real `hf_hub 1.28.0` in the shipped image, offline, both ref states |
| Head path | real cache, deliberately broken ref, real `start.sh` |
| **Worker path** | real worker over ssh, rsync branch, broken ref |
| Idempotency | re-run on both paths, silent |
| Edge cases | echo/CRLF/trailing-space/clean/3×/whitespace-only/missing |

Not covered: a full vLLM serving launch with a broken ref. The failure is at offline resolution, before any weight load, and that step is now tested directly against the real library.

jvr0x:
## Verified on a full serving launch

Ran the A/B on the real rig: identical broken ref, identical cache, identical image, only the branch differs.

**Control — `main`, no fix:**

```
huggingface_hub.errors.LocalEntryNotFoundError: Cannot find an appropriate cached
snapshot folder for the specified revision on the local disk and outgoing traffic
has been disabled.
  File ".../huggingface_hub/_snapshot_download.py", line 354, in snapshot_download

[ERR ] Container vllm-fn exited unexpectedly.
EXIT=1
```

Dies at offline resolution, before a single weight shard is read.

**Fix branch, same broken ref:**

```
[WARN] Normalised trailing whitespace in .../models--nvidia--.../refs/main
       (hf_hub 1.28 offline resolution reads this file without .strip(); see issue #36)
...
Loading safetensors checkpoint shards: 100% | 11/11
[ OK ] vLLM is ready and serving on port 8888!
EXIT=0
```

Then a real request against the served model:

```json
{"messages":[{"role":"user","content":"What is 17 times 4? Answer with just the number."}],
 "chat_template_kwargs":{"enable_thinking":false}}

content: '68'   finish_reason: stop   completion_tokens: 3
```

Same input, opposite outcome, one commit apart.

## Full verification matrix

| layer | evidence |
|---|---|
| Root cause | real `hf_hub 1.28.0` in-image, offline: `abc123\n` → `LocalEntryNotFoundError`; `abc123` → resolves |
| **Serving launch** | **control exits 1 pre-load; fixed serves and answers correctly** |
| Head path | real cache, deliberately broken ref, real `start.sh` |
| Worker path | real worker over ssh, rsync branch |
| Idempotency | re-run on both paths, silent |
| Edge cases | echo / CRLF / trailing-space / clean / 3× / whitespace-only / missing |

Cache restored to its pre-test bytes afterwards; verified identical to the backup.

PR #61, opened 2026-09-16
Release the checkpoint's page cache before launch, without root (#35)

### LINKED ISSUE #35: start.sh never drops page caches (README claim); full page cache OOMs weight loading at 73-97% on GB10
The README (Gotchas) states: *"Drop page caches before every launch (start.sh
does this)."* Neither the pre-`c2325b2` `start.sh` nor current `main` contains
any `drop_caches` code — `grep -n drop_caches start.sh` finds nothing. The
step appears to have been lost somewhere between the README and the script.

**Reproduced twice today** on a 2× DGX Spark kit (GB10, 128 GB unified, same
image `vllm/vllm-openai:qwen38-flash-next`):

- Launch 1 (RadixArk ckpt, fp8 KV): container died at 97% of weight loading
  (shard 199/206), shard load rate had decayed from ~4.5 it/s to ~0.2 it/s —
  classic page-cache/unified-memory contention. buff/cache was ~130 GB at the
  time (a 133 GB checkpoint rsync from NAS had just completed into the same
  cache dir).
- Launch 2 (nvidia ckpt): died at 73% (8/11) immediately after the worker
  rsync finished, same signature.

**Fix that worked (no root needed** — the nodes have no passwordless sudo,
which is presumably why the step went missing**):** evict the checkpoint files'
clean pages via `posix_fadvise(POSIX_FADV_DONTNEED)` as the file owner:

```python
import os
for root in ["/home/jeff/.cache/huggingface/hub/models--*"]:
    for dp, _, fns in os.walk(glob_pattern):
        for fn in fns:
            fd = os.open(os.path.join(dp, fn), os.O_RDONLY)
            os.posix_fadvise(fd, 0, 0, os.POSIX_FADV_DONTNEED)
            os.close(fd)
```

buff/cache went 44 GB → 0, available 117 GB, and both launches then completed
normally. Suggest `start.sh` do this best-effort in the pre-launch step (it
needs no privileges, unlike `echo 3 > /proc/sys/vm/drop_caches`), and either
implement or drop the README claim.

jvr0x:
Implements the no-root half of #35, using the approach the reporter proposed.

## Problem

On GB10 the page cache shares one unified-memory pool with the model. A checkpoint left resident by an `hf download`, the rsync to the worker, or a previous launch gets read a second time into the same pool, and weight loading can die partway with `CUDA out of memory` on an otherwise idle box.

`echo 3 > /proc/sys/vm/drop_caches` fixes it and needs root — which these nodes do not have passwordless. That is exactly why the README tells you to do it yourself rather than claiming `start.sh` does it.

## Approach

`posix_fadvise(POSIX_FADV_DONTNEED)` drops the clean pages of files we can open. No privileges, and it touches **only this checkpoint** rather than the system-wide cache.

`files/evict_page_cache.py` walks the snapshot and evicts weight shards only (`.safetensors`/`.bin`/`.pt`/`.gguf`) — configs and tokenizers are kilobytes and the engine wants them cached. `start.sh` runs it at step 4b-2 on the head and inline over ssh on the worker. `EVICT_PAGE_CACHE=false` opts out.

Best-effort throughout: a missing directory, an unopenable file, or a platform without `posix_fadvise` is skipped rather than failing the launch. Evicting the cache is an optimisation; refusing to start over it would be worse than the problem.

## Verification

Measured per-file residency with `mincore`, not system-wide `free` — precise and unaffected by other processes.

**Head:**
```
before:  4.15 GiB cached of 9.32 GiB  (44.5%)
evict:   page cache: released 123.6 GiB across 11 file(s)
after:   0.00 GiB cached of 9.32 GiB  ( 0.0%)
```

**Worker**, over ssh against an rsync-mode fixture:
```
before:  1.71 GiB cached of 2.00 GiB  (85.3%)
after:   0.00 GiB cached of 2.00 GiB  ( 0.0%)
```

**In situ** — full launch with a deliberately warmed cache (checkpoint 100% resident, `buff/cache` 70 GiB):
```
=== Step 4b-2: Release checkpoint page cache ===
  HEAD   page cache: released 123.6 GiB across 11 file(s)
```
`buff/cache` 70 GiB → 3 GiB, available → 113 GiB. Load then completed, `/health` 200, `EXIT=0`, and the served model answered correctly (`23 × 3` → `69`).

**Data integrity** — clean pages only, so this cannot lose data. Confirmed rather than asserted: sha256 of 2 GiB across an eviction is byte-identical.

**Edge cases:** missing dir, no args, unreadable file, non-weight-only dir — all exit 0 or 2-for-usage, none fatal.

## Known limit

Under `NFS_SHARE=true` the worker reads the checkpoint over NFS, so its resident pages are NFS-client cache that a host-side `fadvise` does not reach. The head is covered in both modes; the worker in the default rsync mode (`NFS_SHARE=false` is what ships). Documented in both README and `.env.sample` rather than left as a surprise.

## Scope

New 83-line helper, 48 lines in `start.sh`, plus README and `.env.sample`. No behaviour change when `EVICT_PAGE_CACHE=false`.

Does not close #35 on its own — the README claim that issue also reported was corrected separately; this adds the part that was still missing.


jvr0x:
## Closed the NFS gap — both nodes now covered in both modes

The first commit left a caveat: under `NFS_SHARE=true` the worker has no local checkpoint, so the ssh-based pass walked an empty directory and did nothing. I documented that as a known limit. That was the wrong call — it's solvable.

**`fadvise` works over NFS 4.2 exactly as it does locally.** Measured inside the mount:

```
after read : 3.00 GiB cached (32.2%)
after evict: 0.00 GiB cached ( 0.0%)
```

The only obstacle was that the mount exists solely inside a container — and the repo already has that pattern: `nfs_worker_has_model()` probes the same volume with a throwaway container. The eviction now runs the same way.

Also removed the duplicated heredoc. One script is copied to the worker and serves both nodes in both modes.

## Verified in NFS mode, both caches deliberately warmed

```
=== Step 4b-2: Release checkpoint page cache ===
  HEAD   page cache: released 123.6 GiB across 11 file(s)
  WORKER page cache: released 123.6 GiB across 11 file(s)
```

Residency measured with `mincore` either side:

| | before | after |
|---|---|---|
| Head | 100.0% | **0.0%** |
| Worker (NFS client cache) | 5.00 GiB / 53.7% | **0.00 GiB** |

Then `EXIT=0`, `/health` 200, and the served model answered correctly (`12 × 12` → `144`).

## Full coverage

| | head | worker |
|---|---|---|
| `NFS_SHARE=false` (default) | ✅ local fadvise | ✅ ssh + local fadvise |
| `NFS_SHARE=true` | ✅ local fadvise | ✅ container + NFS fadvise |

All four paths measured, not inferred. README and `.env.sample` updated — the limitation they described no longer exists.

This now implements #35 fully rather than partially.

